PR #36, opened 2020-10-22
add menu to close/minimize multiple windows

linusha:
Hi,

I added some ease-of-use menu items, the screenshots should be self-explanatory.

windows_world_menu
window_menu

Naturally, I am absolutely open to any criticism!

I could also imagine elaborating on this functionality, e.g. "close all other browsers" in browsers, workspaces,... if that is something that would be nice.

I plan on investigating the whole world-menu to get more accustomed to `lively` and maybe fix some things that are broken right now, however I think this change stands for its own, although it touches the world menu.

### DIFF
--- a/lively.components/window.js
+++ b/lively.components/window.js
@@ -114,6 +114,15 @@ export default class Window extends Morph {
           {alias: "reset",  target: w, command: "resize active window", args: {window: this, how: "reset"}},
         ]
       ],
+      [
+        "Window Management", [
+          {alias: "minimize all except this",   target: w, command: "toggle minimize all except active window"},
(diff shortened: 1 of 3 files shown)


JensLincke:
In Lively4, we have also such a useful global window dialog... with dynamic entries for each window: 
image

merryman:
I am fine with the way the state of this feature is at. @linusha are there any more enhancements you want to make or can this be merged for now?

linusha:
@merryman I have some things jotted down on my list that are somewhat related to this feature but not really to the code here (e.g. the proposal of @JensLincke).
I am absolutely happy with merging this now and I might just open a new windows related PR in the next days. Thank you!

PR #47, opened 2020-10-29
do not reset search by clicking in editor

### LINKED ISSUE #40: Search in code jumps does not allow edit at found code
When I search for a specific term in the code opened in the browser, and the search finds the term, I may want to edit in that place. This is however not possible, as clicking back into the editable code closes the search and returns me to the situation I was in before the search. 

This is unexpected behavior, as other text editors do not act in such a way. Also this makes coding somewhat harder.

linusha:
As it is this closes #40.

There are two additional features I would like to add to this, if you agree that they make sense @merryman:

- [ ] I think it would be cool to have an indication of how many search results there are in the widget
- [ ] (optionally: show the current search position in the widget if the above was implemented (e.g. if there are 14 results and my cursor is at the third show 3/17))

I have to play around more debugging and actually looking at the search results that are being passed around, but I think as it is there is no single array of which we could take the length or something similar?

- [ ] add a button to switch between case-sensitive and insensitive search. My understanding right now is that the search is hardcoded to be case-insensitive?

### DIFF
--- a/lively.morphic/text/search.js
+++ b/lively.morphic/text/search.js
@@ -313,7 +313,7 @@ export class SearchWidget extends Morph {
     setTimeout(() => {
       let focusedMorph = world.focusedMorph;
       if (!this.withAllSubmorphsDetect(m => m.isFocused())) {
-          this.cancelSearch();
+          this.cancelSearch(false);
           return;
       }
       if (this.get('searchInput') != focusedMorph &&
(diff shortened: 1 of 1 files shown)


merryman:
Just a little heads up: In order to adjust the styling of the search widget you will want to open http://localhost:9011/worlds/load?file=lively.morphic/styleguides/SystemWidgets.json and adjust the search widget master in there.
Since the widget creates its layout during initalization, the layout does not need to be properly configured on the master component. This is quite handy since grid layout editing is broken right now, so for convenience you might just want to remove the layout from the search widget master. That should make building the new interface quite a bit easier.
Once I have grid layout editing working we can add that layout back into the master, and remove the initialization code from the SearchWidget itself.

If you have question, just hit me up on slack and we can talk over it.

linusha:
This is obsolete because of #228.

PR #57, opened 2020-11-03
change halo mode icon in top bar

### LINKED ISSUE #50: An arrow with tail rather than a tail-less arrow is standard for selection
I recommend changing the icon for standards' sake
image

linusha:
I saved the `.json` file from inside `lively` and then recreated the `.br.json` from the command line with `brotli 1.0.4`.  I hope this is correct :slightly_smiling_face: 

The new Icon is https://fontawesome.com/icons/mouse-pointer?style=solid, so basically the same as proposed in #50.

resolve #50


merryman:
Superb, thanks!

PR #73, opened 2020-11-11
#56 no drops in dev-tools and option in morphic properties menu

### LINKED ISSUE #56: Developer tools should not react to drops
Developer tools should not react to grabbed objects, soit's not possible to push objects in the world into developer tools.

linusha:
Browser, Object Editor, JS Workspace and Inspector do not accept drops anymore.
The morph menu allows to accept/not accept drops for a whole hierarchy of morphs.

Tooltip_072

We briefly discussed a better solution to block drops in all dev-tools and e.g. color-pickers as well. Since we did not come up with a good solution we opted for this, since these tools were the biggest pain points for us.

This closes #56

### DIFF
--- a/lively.ide/js/browser/index.js
+++ b/lively.ide/js/browser/index.js
@@ -1068,8 +1068,8 @@ export default class Browser extends Morph {
       .map(m => ({ string: m.nameInPackage + (m.isLoaded ? '' : ' [not loaded]'), value: m, isListItem: true }));
 
     await this.ui.moduleList.whenRendered();
-    //since list items are not generated at the time of browser creation,
-    //drops need to be deactivated separately
+    // since list items are not generated at the time of browser creation,
+    // drops need to be deactivated separately
     this.ui.moduleList.doNotAcceptDropsForThisAndSubmorphs();
   }
 
--- a/lively.ide/js/browser/index.js
+++ b/lively.ide/js/browser/index.js
@@ -75,10 +75,7 @@ export default class Browser extends Morph {
     // packageName, moduleName, codeEntity, scroll, textPosition like {row: 0, column: 0}
     const browser = browserOrProps instanceof Browser
       ? browserOrProps : new this(browserOrProps);
-    if (!browser.world()) {
-      const window = browser.openInWindow();
-      window.doNotAcceptDropsForThisAndSubmorphs();
-    }
+    if (!browser.world()) browser.openInWindow();
     return browser.browse(browseSpec, optSystemInterface);
   }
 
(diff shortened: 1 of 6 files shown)


lively.morphic/morph.js
```diff
@@ -2236,6 +2236,20 @@ export class Morph {
     evt.hand.dropMorphsOn(this);
   }
 
+  doNotAcceptDropsForThisAndSubmorphs() {
+    this.acceptsDrops = false;
+    this.submorphs.forEach((m) => {
```
merryman, line 2241: Interesting, I was expecting this to not only look at the direct children of the morph in question, but the entire submorph hierarchy, so all children reachable from that morph.
I can imagine a bunch of scenarios where I have to manually go through each layer of morphs manually and disable the drops, which kind of defeats the purpose of this "shortcut"? 
For instance, calling this on a list will still leave the items in that list unchanged (and continue to accept drops), since these are wrapped inside a list container.
Then again, it avoids unintended disabling of drops, since a smaller set of morphs is altered.
Thoughts?

lively.ide/js/browser/index.js
```diff
@@ -77,7 +77,10 @@ export default class Browser extends Morph {
     // packageName, moduleName, codeEntity, scroll, textPosition like {row: 0, column: 0}
     var browser = browserOrProps instanceof Browser ?
       browserOrProps : new this(browserOrProps);
-    if (!browser.world()) browser.openInWindow(); 
+    if (!browser.world()) {
+      let window = browser.openInWindow();
+      window.doNotAcceptDropsForThisAndSubmorphs();
```
merryman, line 82: This probably skips a bunch of containers inside the browser who still continue to accept drops. From the top of my head I can think of the wrapper for the top row of buttons as one example of that.

lively.ide/js/objecteditor/index.js
```diff
@@ -60,6 +60,7 @@ export class ObjectEditor extends Morph {
         evalEnvironment
       });
     }
+    win.doNotAcceptDropsForThisAndSubmorphs();
```
merryman, line 63: Same as with the browser, what happens for instance with the importController? It will happily continue accepting drops. So either more calls of this are needed or we need to adjust the semantics of this method. (See last comment)

merryman (changes requested):
Overall good idea, but I think we need to pour more thoughts into the actual semantics.

lively.morphic/morph.js
linusha: I am not sure if I understand you correctly, but I actually thought we traverse the whole submorph-hierarchy, since we have the recursive call in line 2242?

merryman: Beats me! You are absolutely right. Please use the idiomatic way in lively which is: `this.withAllSubmorphsDo(m => {})`
However then we end up with the other side of the issues I mentioned. For instance opening up a drag and drop game inside a window would break the game.

linusha:
@merryman as we discussed on monday, we rebased and switched to the idiomatic way to traverse the submorph hierarchy.
Therefore I think, if you find the time, that this can be merged :)

merryman (changes requested):

lively.morphic/morph.js
```diff
@@ -2256,6 +2256,14 @@ export class Morph {
     evt.hand.dropMorphsOn(this);
   }
 
+  doNotAcceptDropsForThisAndSubmorphs () {
+    this.withAllSubmorphsDo(m => { m.acceptsDrops = false; debugger; });
```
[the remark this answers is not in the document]
linusha: Aargh, super embarrassing. Sorry!

PR #97, opened 2020-11-18
Use trash icon for CloseHaloItem

SilvanVerhoeven:
In a user study, we found the "Times" icon in a Morph' shalo not to be clear enough about deleting the whole Morph. That's why we replaced the icon with a "Trash" icon.


merryman:
Super interesting. I will go with it!

PR #98, opened 2020-11-18
Hotfix: enable copying in browser

T4rikA:
In the text/document.js row can't be a const, because it is changed in line 1435. This is a hotfix which solves the problem.


merryman:
Thank you, this is something that must have been accidentally introduced by the linter. Keep watching out for other cases like this.

PR #102, opened 2020-11-24
allow deletion of selected morphs with del-key

### LINKED ISSUE #101: Delete button should remove morph when it is selected with halo
When a morph is selected with its halo, it can be removed with the trash bin button. IMO it should also be possible to press the delete button in that scenario to remove the morph.

linusha:
We could not verify that this works as expected on mac, since we do not have access to one. 

This closes #101


merryman:
Works for me!

PR #150, opened 2021-01-02
prompts: add static example method for all prompts

linusha:
This is just a tiny quality-of-life change. Some of the prompt-types already had a static example function that utilized the wrappers in `world`. I added the remaining ones.


merryman:
How lovely! Thanks.

PR #153, opened 2021-01-06
Labels do not accept drops as default anymore

### LINKED ISSUE #108: Labels shouldn't have accept drops enabled
Screen Shot 2020-11-25 at 16 36 27

linusha:
This closes #108

### DIFF
--- a/lively.morphic/text/label.js
+++ b/lively.morphic/text/label.js
@@ -15,6 +15,10 @@ const { h } = vdom;
 export class Label extends Morph {
   static get properties () {
     return {
+      acceptsDrops: {
+        defaultValue: false
+      },
+
       fill: { defaultValue: Color.transparent },
       draggable: { defaultValue: false },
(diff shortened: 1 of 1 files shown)


lively.morphic/text/label.js
```diff
@@ -214,6 +214,7 @@ export class Label extends Morph {
     super(obj.dissoc(props, ['fontMetric']));
     if (fontMetric) { this._fontMetric = fontMetric; }
     this._cachedTextBounds = null;
+    this.acceptsDrops = false;
```
merryman, line 217: Actually the canonical way to achieve this is to set the defaultValue to `false` in the property definition of acceptsDrops.

merryman (changes requested):

linusha:
Okay, makes sense. Does it use the correct way now? :)

merryman (approved):

PR #159, opened 2021-01-12
Comments

frcroth:
Implements a feature for comments on lively morphs. 
- Comment can be added by right clicking on a morph and using the "Add comment" option
- Comments may be viewed in the Comment Browser, accessible via a new top bar button
- With open comment browser, comments are indicated by little icons
- Comments are objects that have a timestamp, an author, a text and a position in the morph
- In the comment browser, comments may be removed, edited or resolved
- the comment browser has an archive that only shows resolved 
- Add some tests for comments

For more  usage details and screenshots, refer to the new README.md


lively.morphic/morph.js
```diff
@@ -2903,7 +2939,7 @@ export class Image extends Morph {
         );
         this.extent = this.naturalExtent.scaleBy(s);
       }],
-      ['resize image to its real image size', async () => this.extent = await this.determineNaturalExtent()],
+      ['resize image to its real image size', () => this.extent = this.naturalExtent],
```
frcroth, line 2942: What's this?

frcroth, line 2942:
```suggestion
       ['resize image to its real image size', async () => this.extent = await this.determineNaturalExtent()],
```

lively.ide/world.js
```diff
@@ -1023,6 +1025,24 @@ export class LivelyWorld extends World {
     //   items.push(['connect...', connectItems]);
     // }
 
+    items.push({ isDivider: true });
+    items.push(['Add comment', async () => {
+      // TODO: maybe use promise functionality instead of if else
```
frcroth, line 1030:
```suggestion
```

lively.halos/morph.js
```diff
@@ -1382,12 +1386,17 @@ class CopyHaloItem extends HaloItem {
...
     if (isMultiSelection) {
       // FIXME! haaaaack
-      const copies = target.selectedMorphs.map(ea => world.addMorph(ea.copy()));
+      const copies = target.selectedMorphs.map(ea => {
+        if (ea.canBeCopied()) {
+          const copy = ea.copy(true);
+          world.addMorph(copy);
+          return copy;
+        }
+      }).filter(copy => copy);
```
merryman, line 1398: take a look at `arr.compact()`

lively.morphic/world-loading.js
```diff
@@ -82,7 +83,7 @@ export async function loadWorld (newWorld, oldWorld, options = {}) {
       window.history.pushState({}, 'lively.next', options.pathForBrowserHistory);
     }
 
-    newWorld.showsUserFlap = showUserFlap;
+    newWorld.showsUserFlap = showUserFlap; // will initialize Comment Browser
```
merryman, line 86: But this was already triggered. Is there a race condition of some sorts?

lively.morphic/serialization.js
```diff
@@ -81,8 +82,27 @@ export async function saveWorldToResource (world = MorphicEnv.default().world, t
...
+  }
+
+  const serializedMorph = serializeMorph(morph);
+
+  morph.comments = cachedComments;
+  if (morph.attributeConnections) {
+    morph.attributeConnections = morph.attributeConnections.concat(cachedConnections);
+  }
+
+  return deserializeMorph(serializedMorph, { migrations, reinitializeIds: true });
```
merryman, line 105: So in principle that is fine, however one thing that has started to bug me is that the Morph class is waaaay to complex already and I am constantly looking for ways to move code into other modules. In fact it is still one of the main points in my agenda to wipe all the code from lively.morphic that is not essential to rendering stuff and moving it to other modules. This is key to ensuring fast loading times when things are deployed in production. So I personally would prefer to keep logic like this out of Morphic and either solve it by enhancing state monitoring inside the comments system itself or start working with commands instead. _Take a look at how the triggering of layout grids of the halo is handled._

merryman (changes requested):
Overall nice work! Let's talk if we can possibly keep code out of lively.morphic.

lively.morphic/world-loading.js
frcroth: What do you mean this was already triggered?

merryman: I meant the initialization of the Comment Browser.
The property at that point as already been set to that value. 
This therefore triggers the initialization of the comment browser just after the user widgets habe been loaded.
So this assignment does not really do anything as far as I can tell.

Paula-Kli: This is the only line where `showUserFlap` is set. If you do not do that, loading a new world results in not having a top bar at all. We just added the comment which does not totally display what we wanted to say.
What we really wanted to say is, that here is the point what triggers to initialize the comment browser. If there is no top bar no comment browser gets initialized.

lively.morphic/serialization.js
Paula-Kli: but this is a method that is called in the whole world and we added logic but did not write a method ourselves.

how would you put that into another module? 

we needed to do this because when a morph is serialized we want to save it with all of its comments but when a morph is copied the comments should not be copied as well since that would probably not make any sense anymore.

lively.morphic/morph.js
```diff
@@ -2509,6 +2525,26 @@ export class Morph {
...
+  async removeComment (commentToRemove) {
+    this.comments = this.comments.filter(comment => !commentToRemove.equals(comment));
+    CommentBrowser.removeCommentForMorph(commentToRemove, this);
+  }
+
+  emptyComments () {
+    this.comments.forEach((comment) => CommentBrowser.removeCommentForMorph(comment, this));
+    this.comments = [];
+  }
+
```
frcroth, line 2547: Since you say that morph should be as concise as possible, this could also be moved to Comment with static methods that take a morph as an argument. However this may make the methods somewhat harder to understand and these methods would then be responsible for the comments field on some foreign morph. What is your opinion on this, @merryman ?

lively.halos/morph.js
Paula-Kli: good idea!

lively.collab/comments/commentMorph.js
```diff
@@ -0,0 +1,360 @@
...
+    this.isInEditMode = true;
+
+    this.disabledButtonAppearance();
+    this.ui.editSaveButton.tooltip = 'Save Comment';
+
+    this.ui.commentTextField.readOnly = false;
+    this.ui.commentTextField.fill = Color.white;
+    this.ui.commentTextField.borderStyle = 'solid';
+    this.ui.commentTextField.focus();
+  }
```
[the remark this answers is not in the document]
Paula-Kli: We are initializing the ui via code because we want to set it do default in different cases. It is a master component but when a button is clicked different icons are removed and altered so to set it back to default we are doing it here in code. How could we do this via master components to get this behavior?

lively.morphic/world-loading.js
merryman: Beats me, sorry I was confused by the git diff.

lively.collab/comments/commentMorph.js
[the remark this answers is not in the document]
merryman: I am picturing different master components for each situation: a default style master component, and other ones for the other cases respectively. I don't really see you altering the submorph collections. And even that can be handled with showing or hiding certain morphs in the different master components.

lively.collab/comments/commentBrowser.js
```diff
@@ -0,0 +1,351 @@
...
+    return this.resolvedCommentContainer.submorphs.reduce((acc, cur) => cur.getCommentCount() + acc, 0);
+  }
+
+  getUnresolvedCommentCount () {
+    return this.commentContainer.submorphs.reduce((acc, cur) => cur.getCommentCount() + acc, 0);
+  }
+
+  updateCommentCountBadge () {
+    const count = this.getUnresolvedCommentCount();
+    let badge = $world.get('lively top bar').get('comment browser button').get('comment count badge');
```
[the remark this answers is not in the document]
frcroth: Note: Needs to be done on the next restoration of SystemIDE

lively.collab/comments/commentBrowser.js
```diff
@@ -0,0 +1,351 @@
...
+
+    if (comment.isResolved()) {
+      groupDictionary = this.resolvedCommentGroups;
+      commentContainer = this.resolvedCommentContainer;
+    }
+
+    if (!(morph.id in groupDictionary)) {
+      const commentGroupMorph = await resource('part://CommentComponents/comment group morph master').read();
+      await commentGroupMorph.initialize(morph);
+      groupDictionary[morph.id] = commentGroupMorph;
```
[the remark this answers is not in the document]
frcroth: Should this be done with __after_deserialize:: and __additionally_serialize__ or during general activity with something like a decorator on WeakMap?

lively.morphic/morph.js
frcroth: Fine for now

lively.morphic/serialization.js
merryman: For the time being that is fine. Let's push that over to the next refactoring.

merryman (approved):

PR #175, opened 2021-02-10
Add a method so the latest element on the undo stack can be deleted

linusha:
We needed this to accomplish correct undo behavior in a more complex scenario.

Short story is that we check *after* an action was completed if this was an allowed action and revert it manually if this was not the case. This nevertheless created an corresponding undo, since the beginning of the process is the same with legal actions. Therefore we wanted to remove these.

We see that this method needs to be used with care and are open to discuss alternatives/give a more in-depth explanation of what happens in our application if you have strong concerns about this.

### DIFF
--- a/lively.morphic/undo.js
+++ b/lively.morphic/undo.js
@@ -65,7 +65,8 @@ class Undo {
   toString () {
     const { name, changes, no } = this;
     const isRecording = this.isRecording();
-    const changesString = !changes.length ? 'no changes'
+    const changesString = !changes.length
+      ? 'no changes'
       : '\n  ' + changes.map(({ selector, args, prop, value, target }) =>
         selector
           ? `${target}.${selector}(${args.map(printArg)})`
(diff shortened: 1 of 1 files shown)


lively.morphic/undo.js
```diff
@@ -159,6 +160,13 @@ export class UndoManager {
...
 
+  removeLatestUndo () {
+    this.undoStop();
+    const undo = this.undos.pop();
+    arr.remove(this.grouping.current, undo);
+    return undo;
+  }
+
   undo () {
     this.undoStop();
```
merryman, line 171: Can you replace these 2 lines by
` const undo = this.removeLatestUndo();`

merryman (changes requested):

lively.morphic/undo.js
linusha: Just to make sure that I understand you correctly (I sometimes get really confused by the way GitHub formats these, sorry) - we are talking about the implementation of the `undo` method?
If so, will do!

merryman: Yes!

linusha: Done :)

merryman (approved):

Issue #160, opened 2021-01-12
Incorrect article


frcroth:
When a morph is created without a name, its name is "a"/"an" classname. The distinction between a and an is performed based on whether the word starts with a vowel. Unfortunately this is not correct, as this is not the general rule in English. For example "Uniform" is written with a vowel, but the correct article is "a".
The `articlize` method from `articles` (https://jspm.dev/npm:articles@0.2.2!cjs) can do this.

merryman:
Sounds like a nice improvement. Can you prepare a PR that suggests a fix?

frcroth:
> Sounds like a nice improvement. Can you prepare a PR that suggests a fix?

See #189

frcroth:
Closed with #189

PR #189, opened 2021-02-23
Use correct article for default name

frcroth:
See #160 for context.
Since the linter has not yet run on strings.js a lot of additional style changes were added in this commit. The changes I wanted to cause:

- Add articles dependency
- Add findArticle and articlize for strings.js (See line 121 following)
- Use findArticle to use the correct article on default name in morph

### DIFF
--- a/lively.lang/package.json
+++ b/lively.lang/package.json
@@ -29,11 +29,9 @@
       "child_process": {
         "node": "@node/child_process",
         "~node": "@empty"
-      }
+      },
+      "articles": "https://jspm.dev/npm:articles@0.2.2"
     }
-  },
-  "dependencies": {
-    "articles": "^0.2.2"
   },
   "devDependencies": {
     "babel-core": "^6.16.0",
--- a/lively.lang/package.json
+++ b/lively.lang/package.json
@@ -32,7 +32,9 @@
       }
     }
   },
-  "dependencies": {},
+  "dependencies": {
+    "articles": "^0.2.2"
+  },
   "devDependencies": {
     "babel-core": "^6.16.0",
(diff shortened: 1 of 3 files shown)


lively.lang/package.json
```diff
@@ -32,7 +32,9 @@
       }
     }
   },
-  "dependencies": {},
+  "dependencies": {
+    "articles": "^0.2.2"
```
merryman, line 36: Is this package in ES6 module syntax? If not, this will cause lively bundles to fail, since our bundler only supports ESM format.

merryman (changes requested):

lively.lang/package.json
frcroth: I don't think it is: https://jspm.dev/npm:articles@0.2.2!cjs. How would I check this? install.sh ran without problem

merryman: Well usually by inspecting the source files that are pulled by npm. Since npm performs a bunch of preprocessing of scripts, there is no easy way of telling aside from directly inspecting the pulled source files.
Alternatively you can just fetch the package via jspm, then you are on the safe side since jspm transpiles everything tp esm like magic.

frcroth: In the node_modules folder the only code file starts with `(function() {` and ends with  `}).call(this); ` so I don't think it is an ES module. So should I add the dependency as a link on top of lively.lang?

merryman: Yes, but putting the paths inside the dependencies wont work, you need to add a mapping like so: 
```
"systemjs": {
    "map": {
      "articles": "https://jspm.dev/npm:articles@0.2.2"
    }
  },

```

merryman (approved):

PR #207, opened 2021-03-17
Fix backspace shift printing input

### LINKED ISSUE #26: Shift + Backspace prints "input"
Not sure if intended behavior: If I use Shift and Backspace together, the string "input" is printed to the active editor.  
Is this intended? Sometimes I'm writing something with shift and using backspace, which results in this behavior, which is not common to other editors.

frcroth:
Closes #26 
While this does not address the core of the problem, it does resolve that issue.

### DIFF
--- a/lively.morphic/config.js
+++ b/lively.morphic/config.js
@@ -208,6 +208,7 @@ const config = {
       { keys: { win: 'Ctrl-Shift-U', mac: 'Meta-Shift-U' }, command: 'undefine variable' },
 
       { keys: 'Backspace', command: 'delete backwards' },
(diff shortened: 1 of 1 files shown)


merryman:
It is actually part of the solution, let me explain: In fact the single backspace event just never enters the system with key pressed. So even if we are to correctly sort out the delteBackSpace InputEvent that is firing in addition to the Backspace event, there is still the Shift key which makes all of this fire as Shift-* events. So still no deletion happening. What you are doing is exactly what needs to be done in addition to that. Coincidentally it also resolves the issue  because it prevents the deleteBackSpace InputEvent from entering the system by proactively changing the content of the textMorph. I am not a 100% sure what is happening here but I think this is save to marge as is.

PR #235, opened 2021-03-31
top bar: finish shape creation in targets

linusha:
This fixes a problem with the new `attachToTarget` of the topbar:
When creating a new shape by dragging on some other morph than the world as a target, the tooltip that is created while shaping the new morph stayed there forever unless one clicked one time directly after finishing the drag for the shape.

image

This fixes that and it now behaves like expected.
It is likely, that this is not the cleanest solution to the problem. I believe the cause of this is some minor difference in how the events are triggered/how they relate between morphs and the world, especially `onDragEnd` and `onMouseUp`. However, I could not find any problems with this solution when playing around in the world.


linusha:
@merryman this is the changed code snippet:

```javascript
// called from onDragEnd
  finishDragSelectionIfNeeded (evt) {
    if (this._morphSelection) {
      const world = this.world();
      this._morphSelection.fadeOut(200);
      obj.values(this._selectedMorphs).map(m => m.remove());
      world.showHaloForSelection(
        Object.keys(this._selectedMorphs)
          .map(id => world.getMorphWithId(id))
      );
      this._selectedMorphs = {};
      this._morphSelection = null;
    }
    this.handleShapeCreation(evt);
  }
```

I added the last call to `handleShapeCreation`.

merryman:
To be honest I have trouble reproducing the bug you described above on my local system. I will first check if I messed up and forget to push any changes. After that I will decide on merging this. Thank you in any case.

linusha:
> To be honest I have trouble reproducing the bug you described above on my local system. I will first check if I messed up and forget to push any changes. After that I will decide on merging this. Thank you in any case.

For what its worth: @frcroth encountered the same problem.

frcroth:
> 
> 
> To be honest I have trouble reproducing the bug you described above on my local system. I will first check if I messed up and forget to push any changes. After that I will decide on merging this. Thank you in any case.

@merryman This still seems to be a problem for us. Did you find anything unpushed?

merryman:
I just did a fresh install of lively.next and could not reproduce the problem. I did encounter the issue @linusha described in #237 but that is not part of this PR is it?

linusha:
I have no idea what happened initially, but with the current master I cannot reproduce it neither anymore.
I'll therefore close it. For #237 I have proposed a fix in the ticket, I can open a PR if you think that's a valid solution.

PR #253, opened 2021-04-17
Expanding properties/commands in JS browser

### LINKED ISSUE #210: Code entity tree in browser should be able to expand objects
In the right panel that shows classes and methods, it would be nice to also expand other things for example properties and commands, that sometimes are quite long and currently make up only one element.

linusha:
Closes #210 

getters for properties and commands as well as array declarations are now expandable in the browser.

Tests could be another useful addition, they provide no declarations at all in the browser right now, but I am not able to tackle that right now. I wrote it down however. 

---

I handled one case for ...generateUnfolded in Morph explicitly, I do not know if there are many more like it?  (@merryman?)

### DIFF
--- a/lively.ast/lib/code-categorizer.js
+++ b/lively.ast/lib/code-categorizer.js
@@ -114,18 +114,21 @@ function parsePropertiesMethod (node, parent, type) {
     node,
     name: node.key.name
   };
-  const nodes = [propertiesNode];
-  const children = [];
-  propertiesNode.node.value.body.body[0].argument.properties.forEach(property => {
-    children.push({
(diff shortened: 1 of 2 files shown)


merryman (changes requested):
Aside from these two issues, this looks good to merge.

linusha:
Yes, good call!
I would hope the way it is now if we fail when doing the more fine granular parsing, we will just return the initial node without the children, which is equivalent to the base case from the calling method. In these cases the behavior is therefore the same as it is now.

merryman (approved):

PR #254, opened 2021-04-22
morph: signal onAbandon at the beginning of abandon

linusha:
Just a small signal in order to do clean up somewhere else before the morph is abandoned since at the beginning of the method the morph is still there. We need this in order to clean up editor elements when a Morph that is part of the currently open Interactive gets removed via its Halos.


frcroth:
Maybe put it beneath the comment?

linusha:
> Maybe put it beneath the comment?

Done

PR #279, opened 2021-05-27
Comment fixes

Paula-Kli:


lively.ide/world.js
```diff
@@ -1002,9 +1002,11 @@ export class LivelyWorld extends World {
       const commentText = await $world.prompt('Enter comment');
       if (commentText) {
         let relativePosition = pt(0, 0);
-        if (evt) {
-          const xRelative = self.localize(evt.position).x / self.width;
-          const yRelative = self.localize(evt.position).y / self.height;
+        if (evt && !evt.halo) {
+          const evtPosition = pt(evt.domEvt.pageX, evt.domEvt.pageY);
```
merryman, line 1006: That is what evt.position is also based on. Is there a reason you bypass this property directly?

Paula-Kli: Because in my case this only helds an error message:
image
When trying to directly access `evt.position` I get this error
image
and no comment is saved

merryman: That is a bug, and should be fixed

merryman: Should be fixed now f881a7d. Please verify if this works for you.

Paula-Kli: yes this fixed the issue for me! thanks!

merryman (approved):

PR #289, opened 2021-06-03
Feature/resizeable panel

SilvanVerhoeven:
Introduces resizable panels to lively.next


merryman (commented):
Can you add some test cases here that illustrate how this is to be used?

merryman:
@SilvanVerhoeven Thank you for the tests, this explains a lot. Did you have a look at: lively.components/resizers.js ? Can you tell me what exactly this kind of resizer allows us to do that these other resizers dont?

PR #292, opened 2021-06-03
Trigger relayout of NumberWidget

### LINKED ISSUE #290: Updating the maximum of the numberWidget does not update everything
When updating the max value of the NumberWidget the max of the actual input field stays the same. The update of this max number is only called through relayout(false).

Paula-Kli:
Closes #290

### DIFF
--- a/lively.ide/value-widgets.js
+++ b/lively.ide/value-widgets.js
@@ -417,8 +417,20 @@ export class NumberWidget extends Morph {
           this.relayout(false);
         }
       },
-      min: { defaultValue: -Infinity },
-      max: { defaultValue: Infinity },
+      min: {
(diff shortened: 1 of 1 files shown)


merryman:
This bodes well with serialization? A simple `numberWidget.copy()` should be able to verify that.

Paula-Kli:
> This bodes well with serialization? A simple `numberWidget.copy()` should be able to verify that.

yes tried that because I was worried, that the setter would be triggered on deserialization and could break things but it works ^^

merryman:
OK, then this is fine for now. Reminds me, let's add a test to lively.components soon that tests for all widgets to be copyable.

PR #302, opened 2021-06-22
tabs: use green for status message upon good renaming of tabs status

linusha:
Just using red for the status message on fail leads to a red check mark being shown which I think is confusing. Changing to a cross (like the custom `error` function in `qinoq`) requires some more code (3 lines or so). My proposal would be to use this for now and later on make it so that one can specify directly in the `setStatusMessage` which icon should be displayed. I'd prefer this over duplicating the code of our custom error message. 

Screenshot (2)


merryman:
Thanks! I will be building a revised version of the status message morphs soon. So part of That can be the aforementioned api change for custom icons. Will keep you posted.

PR #303, opened 2021-06-22
prompt: add option to select input upon opening

linusha:
Also preselects the already input text when changing the content of a label or the image URL over the morph menu.

This was feedback from the user testing on Friday, that it is cumbersome to change those values.

### DIFF
--- a/lively.ide/world.js
+++ b/lively.ide/world.js
@@ -732,7 +732,7 @@ export class LivelyWorld extends World {
     return this.openPrompt(new InformPrompt({ label, ...opts }), opts);
   }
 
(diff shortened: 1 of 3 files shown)


merryman (approved):

lively.ide/world.js
```diff
@@ -732,7 +732,7 @@ export class LivelyWorld extends World {
     return this.openPrompt(new InformPrompt({ label, ...opts }), opts);
   }
 
-  prompt (label, opts = { requester: null, input: '', historyId: null, useLastInput: false }) {
+  prompt (label, opts = { requester: null, input: '', historyId: null, useLastInput: false }, selectInput = false) {
```
frcroth, line 735: Shouldn't this be an option?

PR #304, opened 2021-06-22
promtp: use inputline prompt for preselection

linusha:
@merryman I was a bit trigger happy with the last one.
@frcroth made a good point about this option being better suited to the opts of the input line as opposed to being their own argument.

I'd agree with this assessment. 
If you want, we could probably get rid of the previous merge commit to keep the history cleaner...


frcroth (approved):

merryman (approved):

merryman:
@frcroth good catch, I agree.

PR #352, opened 2022-01-10
Allow setting different borderRadius per corner

linusha:
E.g. the top-corner can have a different radius than the bottom-right one.


lively.morphic/morph.js
```diff
@@ -639,31 +635,31 @@ export class Morph {
...
             value = {
-              left: value.left(),
-              right: value.right(),
-              top: value.top(),
-              bottom: value.bottom()
+              topLeft: value.x,
+              topRight: value.y,
+              bottomRight: value.width,
+              bottomLeft: value.height
             };
```
merryman, line 656: Well this is not longer really supported I guess, so better just remove this one.

merryman (changes requested):

merryman (approved):

PR #361, opened 2022-02-21
Compare colors with delta instead of ===

T4rikA:
In the color.js equals implementation, colors were compared with ===, this caused errors because of the rounding of floats.
It was changed to a delta

### DIFF
--- a/lively.graphics/color.js
+++ b/lively.graphics/color.js
@@ -267,7 +267,7 @@ export class Color {
   equals (other) {
     if (!other) return false;
     const delta = 0.00000001;
-    return this.r - other.r < delta && this.g - other.g < delta && this.b - other.b < delta && this.a - other.a < delta;
+    return Math.abs(this.r - other.r) < delta && Math.abs(this.g - other.g) < delta && Math.abs(this.b - other.b) < delta && Math.abs(this.a - other.a) < delta;
   }
 
   // -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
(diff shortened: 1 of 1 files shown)


linusha:
Could you sketch some scenarios out for me that lead to two "equal" colors being unequal due to rounding errors?

Currently, this solution does seem to be a bit sketchy to me. I would expect `equals` to only return true if two colors are truly equal, e.g. `Color.red.equals(Color.red)`. This is not the case when there are different numbers, regardless of how they came to be, i.e., rounding errors in this case.

So, depending on how this scenario comes to be, I would argue that we either should tackle the places in the code that create this scenario, or make this a separate function instead of changing `equals`.

T4rikA:
Here is one scenario:
Create a new morph
Set the fill of the morph to the first color in the material design pallet
Copy the morph
Print the fill of both morphs

image

In this scenario the equal method would say the colors do not equal. This is because of a rounding error in the decimal representation of the color.

Maybe changing the decimal representation to 0-255 would solve this issue.

linusha:
> Here is one scenario: Create a new morph Set the fill of the morph to the first color in the material design pallet Copy the morph Print the fill of both morphs
> 
> image
> 
> In this scenario the equal method would say the colors do not equal. This is because of a rounding error in the decimal representation of the color.

I pushed some code that should stop this from ever happening by truncating the decimal places in the constructor of Color.
Functionally, this is more or less the same as your suggestion I think, but imo it is a bit easier to understand, especially since one does not need to wonder about the delta stuff in equals. 
Could you check if this solved your problem? :slightly_smiling_face: I quickly tested it with the above scenario of yours and it worked there...

T4rikA:
Just checked it an it works in my use case, thanks!

PR #383, opened 2022-06-15
Serializer modules decoupling

merryman:
This removes the top level dependency onto `lively.modules` within 'lively.morphic/serializer.js' and replaces it with a dependency injection mechanism.

### DIFF
--- a/lively.morphic/serialization.js
+++ b/lively.morphic/serialization.js
@@ -64,7 +64,7 @@ export function deserializeMorph (idAndSnapshot, options = {}) {
 /**
  * Creates a deep copy of a morph.
  * If desired, the copy can also carry over comments that have been assigned to the morphs.
- * Fixme: The commends should be removed from the morphs entirely, since they only make
+ * FIXME: The comments should be removed from the morphs entirely, since they only make
  * sense in the context of the IDE world.
  * @param { Morph } morph - The morph to copy.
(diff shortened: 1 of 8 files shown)


lively.morphic/serialization.js
```diff
@@ -23,68 +38,38 @@ function normalizeOptions (options) {
...
-    i.label = 'Uploading...';
-    await i.whenRendered();
-    return toResource.writeJson(snap);
-  } finally { i && i.remove(); }
-}
-
+/**
+ * Creates a deep copy of a morph.
+ * If desired, the copy can also carry over comments that have been assigned to the morphs.
+ * Fixme: The commends should be removed from the morphs entirely, since they only make
```
linusha, line 67:
```suggestion
 * FIXME: The comments should be removed from the morphs entirely, since they only make
```

lively.morphic/serialization.js
```diff
@@ -111,109 +96,15 @@ export function copyMorph (morph, realCopy = false) {
...
-    migrations,
-    ...options
-  });
-}
-
-// -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
-// helper
-async function resourceToJSON (currentResource, base) {
+/**
+ * Convertes a resource handle pointing to a directory recursively
```
linusha, line 100:
```suggestion
 * Converts a resource handle pointing to a directory recursively
```

PR #387, opened 2022-06-21
Fix source transform tests

merryman:
This will fix some issues with the destructuring assignment source transformation and also fix the tests of the source-transform package.


linusha (approved):
Tests work again on my machine.

PR #398, opened 2022-06-30
Improve behavior upon creation/deletion of packages in browser

linusha:
In the best case we merge this after merging #366.

Only the two commits that change files in `lively.ide` are important for this.

### DIFF
--- a/lively.ide/js/browser/index.js
+++ b/lively.ide/js/browser/index.js
@@ -972,7 +972,7 @@ export class BrowserModel extends ViewModel {
   whenModuleUpdated () { return this.state.moduleUpdateInProgress || Promise.resolve(); }
 
   async selectPackageNamed (pName) {
-    const p = pName ? await this.systemInterface.getPackage(pName) : null;
+    const p = pName ? await this.systemInterface.getPackage(pName) : await this.systemInterface.getPackage('lively.morphic');
     const columnView = this.ui.columnView;
     const td = columnView.treeData;
     await columnView.setExpandedPath(n => {
--- a/lively.ide/js/browser/index.js
+++ b/lively.ide/js/browser/index.js
@@ -969,7 +969,10 @@ export class BrowserModel extends ViewModel {
   }
 
   whenPackageUpdated () { return this.state.packageUpdateInProgress || Promise.resolve(); }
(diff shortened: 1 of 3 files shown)


lively.ide/js/browser/index.js
```diff
@@ -1030,6 +1039,12 @@ export class BrowserModel extends ViewModel {
     return moduleLists.map(list => list.items.map(m => m.value)).flat();
   }
 
+  getDisplayedPackageNodes () {
+    const columnView = this.ui.columnView;
+    const packageList = columnView.submorphs[0];
+    return packageList.items.map(m => m.value).flat();
```
merryman, line 1045: How about: `return this.ui.columnView.treeData.root.subNodes`?;

lively.ide/world.js
```diff
@@ -713,7 +713,7 @@ export class LivelyWorld extends World {
       if (typeof opts.customize === 'function') { opts.customize(promptMorph); }
 
       if (opts.animated) {
-        if (this.previousPrompt && this.previousPrompt.world()) {
+        if (this.previousPrompt && this.previousPrompt.world() && this.previousPrompt.transitionTo) {
```
merryman, line 716: I guess that's fine, but prompt transitioning is broken anyways.
So let's write a note to fix that at some point in the future.

merryman (changes requested):

lively.ide/world.js
linusha: Alright, I'll put it on the list.

lively.ide/js/browser/index.js
```diff
@@ -1125,13 +1139,23 @@ export class BrowserModel extends ViewModel {
...
+    this.ui.sourceEditor.opacity = 0.7;
+    this.ui.sourceEditor.readOnly = true;
+    this.updateSource('');
+    this.ui.metaInfoText.showInactive();
+  }
+
+  reactivateEditor () {
+    this.ui.sourceEditor.opacity = 1;
+    this.ui.sourceEditor.readOnly = false;
+    this.ui.metaInfoText.showInactive();
```
merryman, line 1156: Just a nitpick, but same as earlier, I prefer a top level: `const { sourceEditor, metaInfoText } = this.ui;` instead of having countless `this.ui` refs in the code.

merryman (approved):

PR #404, opened 2022-07-01
Fix tests in lively modules

merryman:


lively.modules/tests/package-test.js
```diff
@@ -467,6 +467,7 @@ describe('package registry', () => {
 
   describe('lookup', () => {
     it('from packageBaseDirs', async () => {
+      // semver somehow does not correctly check the versions any more
```
linusha, line 470: Should this still be here? :see_no_evil: :thinking:

lively.modules/tests/module-test.js
```diff
@@ -147,7 +147,12 @@ describe('module loading', () => {
...
       await importPackage(S, testDir);
       let m = await module(S, testDir + 'file1.js');
-      await m.removeImports([{ local: 'y' }]);
+      expect(await m.source()).equals("import { y } from './file2.js'; export var x = y + 1;");
+      try {
+        await m.removeImports([{ local: 'y' }]);
+      } catch (err) {
+        // causes an error because the code is not longer valid
+      }
       expect(await m.source()).equals(' export var x = y + 1;');
```
linusha, line 156: I am a bit confused - why do we trigger the catch clause **and** the import statements are removed from the code?

Additionally, I believe this is a nice use case for https://www.chaijs.com/api/bdd/#method_throw?

linusha (approved):

PR #407, opened 2022-07-05
🧰: Reenable highlighting of TODO and FIXME comments

linusha:
This solution is improved over the older variant that we had and is taken from here: https://discuss.codemirror.net/t/highlighting-fixme/171/3.
When we update mode.js from a newer CodeMirror version, we need to think of this and reconcile this part manually.

### DIFF
--- a/lively.ide/js/mode.js
+++ b/lively.ide/js/mode.js
@@ -128,9 +128,9 @@ defineMode('javascript', function crea (config, parserConfig) {
     } else if (ch == '/') {
       if (stream.eat('*')) {
         state.tokenize = tokenComment;
-        return tokenComment(stream, state);
+        return ret('comment', 'comment');
       } else if (stream.eat('/')) {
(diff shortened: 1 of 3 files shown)


lively.ide/js/mode.js
```diff
@@ -128,9 +128,11 @@ defineMode('javascript', function crea (config, parserConfig) {
     } else if (ch == '/') {
       if (stream.eat('*')) {
         state.tokenize = tokenComment;
-        return tokenComment(stream, state);
+        return ret('comment', 'comment');
+        // return tokenComment(stream, state);
       } else if (stream.eat('/')) {
-        stream.skipToEnd();
+        // stream.skipToEnd();
+        state.tokenize = tokenLineComment;
```
merryman, line 135: I guess we can remove those comments for good.

merryman (changes requested):

PR #410, opened 2022-07-05
Fix/lively system interface tests

merryman:
Fixes the tests for `lively-system-interface`.

### DIFF
--- a/lively-system-interface/tests/http-interface-test.js
+++ b/lively-system-interface/tests/http-interface-test.js
@@ -3,21 +3,21 @@ import { expect } from 'mocha-es6';
 import { startServer } from './http-server-for-interface.js';
 import { serverInterfaceFor } from '../index.js';
 
+let server, system;
 describe('http lively-system-interface', function () {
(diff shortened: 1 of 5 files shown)


lively-system-interface/tests/http-interface-test.js
```diff
@@ -3,21 +3,21 @@ import { expect } from 'mocha-es6';
...
   before(async () => {
-    server = (await startServer('/lively-tester', 3011)).server;
+    ({ server } = await startServer('/lively-tester', 3011));
     system = serverInterfaceFor('http://localhost:3011/lively-tester');
   });
 
   after(async () => {
-    await server.kill();
+    server.kill();
   });
```
linusha, line 17: This does not need to be `async` anymore, does it?

linusha (changes requested):
Could you amend the updated installer bundle into the commit that changed the installers `packages-config`?

linusha (approved):

linusha (approved):

PR #422, opened 2022-07-08
No nav when unsaved

linusha:


lively.components/muller-columns.cp.js
```diff
@@ -86,6 +87,10 @@ export class MullerColumnViewModel extends ViewModel {
     arr.last(this.lists.filter(m => !!m.selection)).focus();
   }
 
+  get listNavigationProhibited () {
+    return this.view.owner.listNavigationProhibited;
```
merryman, line 91: While this is not as bad as `owner.owner` there is still an implicit assumption about this component being embedded in a certain context and it will break if that is not the case.
How about handling this via the treeData? The treeData can support arbitrary custom attributes which the enclosing morph/context can set accordingly.
So something like `this.treeData.listNavigationProhibited` ....

merryman (changes requested):

linusha:
@merryman please do **not** merge this in case of approval, I need to fix a typo in a commit message before :see_no_evil:

merryman (approved):

merryman (approved):

PR #388, opened 2022-06-21
New component implementation

merryman:
Towards a new internal "morph free" representation of master component policies.

PR #424, opened 2022-07-14
Fix/code search buttons

merryman:
This fixes some style and alignment issues with the code search. Can be merged once we finish #388
buttons preview

### DIFF
--- a/lively.ide/code-search.cp.js
+++ b/lively.ide/code-search.cp.js
@@ -105,6 +105,7 @@ const CodeSearch = component({
       name: 'holder',
       layout: new TilingLayout({
         axisAlign: 'center',
+        align: 'center',
         axis: 'column'
       }),
       submorphs: [
@@ -136,7 +137,7 @@ const CodeSearch = component({
             wrapSubmorphs: false,
             hugContentsHorizontally: true
           }),
-          extent: pt(125.7, 27),
+          extent: pt(125.7, 22),
           borderColor: Color.gray,
           viewModel: {
             openListInWorld: true,
@@ -148,9 +149,7 @@ const CodeSearch = component({
           }]
         })]
     })]
-
-  }
-  )]
+  })]
 })
  ;
 
--- a/lively.ide/code-search.cp.js
+++ b/lively.ide/code-search.cp.js
@@ -105,12 +105,14 @@ const CodeSearch = component({
       name: 'holder',
       layout: new TilingLayout({
         axisAlign: 'center',
+        spacing: 2,
         align: 'center',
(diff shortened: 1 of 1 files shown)


linusha:
@merryman two general "process thoughts":

1. For UI changes, new tools, etc. I think we should try to make a habit of providing a screenshot. In an ideal world the screenshot would probably indicate what *exactly* changed, or even a before/after pair of pictures, but that seems to be a bit too much for small changes like this.
2. I know you were against this in the past, but this again would imo be a prime example for a PR that should not go against main. My argument goes like this: 

- Targeting [new-component-implementation](https://github.com/LivelyKernel/lively.next/tree/new-component-implementation) would make it clear, that this PR is dependent on #388.
- It makes it way easier to see with a glance which commits are actually part of this PR, since the majority of commits here actually belong to #388, which will become apparent once that is merged and we rebase here, but still...

Nothing major or dealbreaking of course, but I wanted to jump on the opportunity to defend this approach once again :smile: 
In case your argument against PRs that do not target main is really only about discoverability, I can [automate a Label](https://selleo.com/til/posts/iumvmfrkhn-automate-pull-request-labels-via-github-actions)  that makes it clear when a PR does not target main.

merryman:
> @merryman two general "process thoughts":
> 
> 1. For UI changes, new tools, etc. I think we should try to make a habit of providing a screenshot. In an ideal world the screenshot would probably indicate what _exactly_ changed, or even a before/after pair of pictures, but that seems to be a bit too much for small changes like this.

Agreed, I updated the PR accordingly.

> 2. I know you were against this in the past, but this again would imo be a prime example for a PR that should not go against main. My argument goes like this:
> 
> * Targeting [new-component-implementation](https://github.com/LivelyKernel/lively.next/tree/new-component-implementation) would make it clear, that this PR is dependent on [New component implementation #388](https://github.com/LivelyKernel/lively.next/pull/388).
> * It makes it way easier to see with a glance which commits are actually part of this PR, since the majority of commits here actually belong to [New component implementation #388](https://github.com/LivelyKernel/lively.next/pull/388), which will become apparent once that is merged and we rebase here, but still...
> 
> Nothing major or dealbreaking of course, but I wanted to jump on the opportunity to defend this approach once again 😄 In case your argument against PRs that do not target main is really only about discoverability, I can [automate a Label](https://selleo.com/til/posts/iumvmfrkhn-automate-pull-request-labels-via-github-actions) that makes it clear when a PR does not target main.

Okay, once we have this automatic label in place I am ready to take another stab at merging into feature branches instead of main. I will then update the PR as you suggested.

PR #442, opened 2022-08-31
New bootstrap bundles

merryman:
This can be reviewed once #388 is merged.
Here we basically introduce a new implementation of landing and loading page in lively.next, which is based on the declarative component definitions.
We also provide build scripts for each, which can be run from the shell.
@linusha We should use this as the foundation for a new Github Actions Check that verifies that after changes to the system core, the loading and landing pages have been rebuilt.


linusha:
Since we need to rebase this anyways, might as well fix the typo in ffba0d9e30f9adb9d740a51dc04466313f9a201b

linusha:
> We should use this as the foundation for a new Github Actions Check that verifies that after changes to the system core, the loading and landing pages have been rebuilt.

1. What exactly do you refer to as the system core?
2. In 172319b2f62797a56cfc0e79555b87d5a6f544a2 it looks like the artifacts are everything that is inside of the `landing-page` and `loading-screen`dir inside of the freezer, is that correct?

merryman:
> > We should use this as the foundation for a new Github Actions Check that verifies that after changes to the system core, the loading and landing pages have been rebuilt.
> 
> 1. What exactly do you refer to as the system core?

That's basically all the modules that are included in the bootstrap bundles. It is most of the packages inside the lively.next repo, except for some of the experimental ones such as `lively.collab` etc. But I will make sure the dependencies are properly declared in the `package.json` of `lively.freezer`. Basically the check should work in the same way as for `flatn` and `lively.installer`.

> 2. In [172319b](https://github.com/LivelyKernel/lively.next/commit/172319b2f62797a56cfc0e79555b87d5a6f544a2) it looks like the artifacts are everything that is inside of the `landing-page` and `loading-screen`dir inside of the freezer, is that correct?

Yes.

lively.freezer/src/bundler.js
```diff
@@ -480,22 +480,39 @@ export default class LivelyRollup {
...
-        return this.resolved[resolutionId(id, importer)] = await this.resolveRelativeImport(importer, id);
+        absolutePath = await this.resolveRelativeImport(importer, id);
+        if (this.belongsToExcludedPackage(absolutePath)) return null;
+        return this.resolved[resolutionId(id, importer)] = absolutePath;
       } catch (err) {
         return null;
       }
     }
 
     if (!id.endsWith('.js') && !isCdnImport) {
```
linusha, line 495: Are we sure this works? Above, I could only find a function with this name and I do not see a variable declared anywhere in this method? :thinking:

linusha (approved):

PR #446, opened 2022-08-31
Fix/system browser meta info

merryman:
Can be reviewed once #442 is merged.
This fixes an issue with the System browser, where upon initial opening, the meta info field would be styled incorrectly.


linusha:
Superseded by #450.

PR #447, opened 2022-08-31
Fix/property side bar bugs

merryman:
Can be reviewed once #442 has been merged.
This fixes three major bugs in the side bar:

1. The opening and closing animation of the flaps was not properly in sync with the sidebars or had weird offsets.
2.  The border control section of the property panel was not handling differently styled borders correctly.
3. Upon the first shape creation, it accidentally caused the colour of the border to be transparent, instead of the default blue.


linusha:
Superseded by #450.

PR #448, opened 2022-09-02
Integration of new renderer without vdom

linusha:
@merryman you should have r/w access to my Project for this [here](https://github.com/users/linusha/projects/1). Just ping me if its not working.


lively.ide/studio/dialogs.cp.js
```diff
@@ -143,20 +143,26 @@ class SaveWorldDialogModel extends ConfirmPromptModel {
...
   name: 'save world dialog',
   extent: pt(470, 320),
+  layout: new VerticalLayout({
+    align: 'center',
+    autoResize: true,
+    direction: 'topToBottom',
+    renderViaCSS: true,
+    resizeSubmorphs: true,
+    spacing: 16
+  }),
```
linusha, line 157: Can we make this a `TilingLayout`? :slightly_smiling_face: Otherwise, stuff will break later on with #547.

lively.morphic/rendering/renderer.js
```diff
@@ -1796,13 +1796,14 @@ export default class Renderer {
...
     const prevParent = textNode.parentNode;
+    textNode.remove();
     this.placeholder.className = 'Text';
-    this.placeholder.appendChild(textNode);
     textNode.style.width = 'max-content';
     textNode.style.height = 'max-content';
     textNode.style.position = 'static';
     if (morph.fixedWidth) textNode.style.width = morph.width + 'px';
     if (morph.fixedHeight) textNode.style.height = morph.height + 'px';
+    this.placeholder.appendChild(textNode);
```
linusha, line 1807: I do not understand this change. According to the commit message, this should make sure that the node gets removed from the placeholder. However, the removal of the placeholder is what happens in line 1811 I'd think?

Explicitly calling remove is also okay I guess, but as I understood it, both versions should behave exactly the same as appending a node that is mounted somewhere else results in the node being removed there.

lively.ide/js/workspace.js
```diff
@@ -27,6 +27,7 @@ export default class Workspace extends Window {
             name: 'editor',
             lineHeight: 1.4,
             readOnly: false,
+            needsDocument: true,
```
linusha, line 30: We can of course say that we always want to set this explicitly, but if you did not change that (I have not yet read all changes), this is not necessary. Since this morph has `readOnly == false`, its text will always be backed by a document, as we need that for interactive editing capabilities.
The flag was introduced for cases where we do not want to set `readOnly = true`, but still need some capabilities that the document provides us. The prime example are interactive trees like the Inspector or the scene graph, where we do not want to interactively edit the content.

I actually remember, that we talked about this in the past and you were not super happy with the solution and wanted to have a third user exposed flag? But I am not sure anymore, sorry :(

lively.morphic/text/morph.js
```diff
@@ -2492,10 +2492,6 @@ export class Text extends Morph {
       .insetBy(this.borderWidth);
     let charBounds = this.charBoundsFromTextPosition(pos);
 
-    // if no line wrapping is enabled we add a little horizontal offset so
-    // that characters at line end are better visible
-    if (!this.lineWrapping) charBounds = charBounds.insetByPt(pt(-20, 0));
-
```
linusha, line 2226: Ah, I see.
I never had problems with this line, but I am also not exactly sure what its purpose was? If you had a problem that is solved now, I would think its okay.

lively.morphic/rendering/renderer.js
```diff
@@ -1680,8 +1680,9 @@ export default class Renderer {
     const horizontalScrollBarVisible = morph.document.width > morph.width;
     const scrollBarOffset = horizontalScrollBarVisible ? morph.scrollbarOffset : pt(0, 0);
     const verticalPaddingOffset = morph.padding.top() + morph.padding.bottom();
-    scrollLayer.firstChild.style.width = Math.max(morph.document.width, morph.width) + 'px';
-    scrollLayer.firstChild.style.height = Math.max(morph.document.height, morph.height) - scrollBarOffset.y + verticalPaddingOffset + 'px';
+    const horizontalPaddingOffset = morph.padding.left() + morph.padding.right();
+    scrollLayer.firstChild.style.width = morph.document.width + horizontalPaddingOffset + 'px';
+    scrollLayer.firstChild.style.height = morph.document.height - scrollBarOffset.y + verticalPaddingOffset + 'px';
```
linusha, line 1685: Does this solve the issue that the bounds of some nodes in the DOM were effectively unbound?

lively.morphic/world.js
```diff
@@ -202,7 +202,7 @@ export class World extends Morph {
 
   onMouseDown (evt) {
     if (evt.state.menu) evt.state.menu.remove();
-    this.onWindowScroll();
+    // this.onWindowScroll();
```
linusha, line 205: Why? :thinking:

lively.morphic/rendering/property-dom-mapping.js
```diff
@@ -11,9 +11,14 @@ import bowser from 'bowser';
...
  */
 export function stylepropsToNode (styleProps, node) {
+  node.style.removeProperty('padding-left');
+  node.style.removeProperty('padding-right');
+  node.style.removeProperty('margin-bottom');
+  node.style.removeProperty('margin-top');
   for (let prop in styleProps) {
     node.style[prop] = styleProps[prop];
   }
+  // FIXME: remove the props NOT mentioned in styleProps but still mentioned in style?
```
linusha, line 21: > ist das richtig das stylepropsToNode nur props hinzufügt oder überschreibt aber keine löscht? Also ist das ein feature oder ein bug?

This is an interesting question!

I had not thought about this previously, which I think is a byproduct of us setting "default values" (i.e., what CSS would default to) quite excessively, so that e.g., a removed fill color seemed to work because it was explicitly overridden (might not be the best example, I hope you catch my drift).

Hovewer, I believe you are right and we should delete unset properties here. I'd propose to just set `node.style = {}` at the beginning of this function? In case this breaks stuff, I think we would have a bug in places where we populate `styleProps`, as I'd expect this to be an idempotent operation.
However, we would still have the possibility to go the hacky route of explicitly deleting properties here, in case tracing problems in the correct solution gets to confusing.

**TL;DR:** I think this should delete all style properties that are set on `node` but are not specified in `styleProps`.

lively.morphic/rendering/renderer.js
```diff
@@ -161,13 +161,17 @@ export default class Renderer {
     }
 
     for (let morph of this.morphsWithStructuralChanges) {
-      this.renderStructuralChanges(morph);
+      morph.withAllSubmorphsDo(m => !this.renderMap.has(m) && this.renderMorph(m));
```
linusha, line 165: This is probably something better suited for a call, but I find this to be dubious...

I assume this is to "make the nodes ready" for any added morphs? In which scenarios was the previous solution problematic? In `renderStructuralChanges`, those are the first two lines:

```
let node = this.getNodeForMorph(morph);
if (!node) node = this.renderMorph(morph);
```
I do not understand what this covers that is not already covered.

lively.morphic/rendering/renderer.js
```diff
@@ -161,13 +161,17 @@ export default class Renderer {
...
     }
 
     for (let morph of this.renderedMorphsWithChanges) {
       this.renderStylingChanges(morph);
     }
 
+    for (let morph of this.morphsWithStructuralChanges) {
+      this.renderStructuralChanges(morph);
+    }
+
```
linusha, line 175: Why does this come after the "styling loop"?

lively.morphic/rendering/renderer.js
```diff
@@ -545,13 +559,20 @@ export default class Renderer {
...
   updateNodeScrollFromMorph (morph) {
-    if (morph.isText && morph.document) return;
-
+    if (morph.clipMode !== 'auto' && morph.clipMode !== 'scroll') return;
     const node = this.getNodeForMorph(morph);
     if (!node) {
       return;
     }
     const { x, y } = morph.scroll;
+    if (morph.isText && morph.document) { // FIXME: maybe check if we are insisting for the document?
```
linusha, line 569: What do you mean by insisting? That we should check if we can remove the second part of the condition?

lively.morphic/rendering/renderer.js
```diff
@@ -1493,7 +1513,7 @@ export default class Renderer {
    */
   renderTextAndAttributes (node, morph) {
     const textNode = node.querySelector(`#${morph.id}textLayer`);
-
+    if (!textNode) return;
```
linusha, line 1517: Just for safety - you have tried searching, expanding snippets, changing font colors, font sizes, copying a text,... after this change? So basically the whole palette of what the TextMorph has to offer?

lively.morphic/rendering/renderer.js
merryman: We can certainly replace
```
morph.withAllSubmorphsDo(m => !this.renderMap.has(m) && this.renderMorph(m))
```
with
```
morph.withAllSubmorphsDo(m => this.renderStructuralChanges(m));
```

However that would be overkill for what we want to achieve here, which is just the "preparation" of nodes. `renderStructuralChanges` does way more than that.

Long story short: Not having all the nodes already initialized before we apply styling changes caused all kinds of issues. One of them was that scrolls did not get set correctly. I will try to review what issues those were in particular later on.

lively.morphic/rendering/renderer.js
merryman: Basically. Discovered no issues so far.

lively.morphic/rendering/renderer.js
merryman: Yes by `morph.needsDocument`, that was my thought.

lively.morphic/rendering/renderer.js
linusha: > However that would be overkill for what we want to achieve here, which is just the "preparation" of nodes. `renderStructuralChanges` does way more than that.

Okay, that makes sense. From this follows that we can remove the node initialization from `renderStructuralChanges()`, doesn't it? Since we made sure to initialize all nodes beforehand.

Nothing that we need to act on super urgently, but my feeling when working on this was that the node-patching etc. is complex enough as it is, and having the general loop quite tidy made developing and especially debugging quite a bit more enjoyable.

lively.ide/js/workspace.js
merryman: This was a temporary change, that fixed the initial problem with workspaces crashing the world when they are deserialized.

lively.morphic/world.js
merryman: This is also a temporary changed that interfered with master component application. I will need to reintroduce that, since its important to check if the scroll of the world is correct every once in a while. We needed this line here in particular for incorrect transform applied to morph that did not take into account the world scroll.

lively.morphic/rendering/renderer.js
merryman: Yes we can certainly refactor to separate the node initialization from the structural change application!

lively.morphic/rendering/renderer.js
merryman: See above conversation.

lively.morphic/text/morph.js
```diff
@@ -13,7 +13,6 @@ import { Range } from './range.js';
 import { eqPosition, lessPosition } from './position.js';
 import KeyHandler from '../events/KeyHandler.js';
 import { UndoManager } from '../undo.js';
-import { TextSearcher } from './search.js';
```
linusha: 🔖 Bookmark for myself, can be safely ignored :slightly_smiling_face: 
Please do not resolve this!

lively.morphic/text/morph.js
merryman: Me neither, but removing it fixed a bunch of things! I will keep this change until I discover some unintended behavior.

lively.morphic/layout.js
```diff
@@ -738,6 +744,7 @@ export class TilingLayout extends Layout {
    * @param { Morph } morph - The layoutable submorph for which to update the bounds for.
    */
   updateBoundsFor (morph) {
+    morph.renderingState.cssLayoutToMeasureWith = false;
```
linusha, line 747:
```suggestion
    morph.renderingState.cssLayoutToMeasureWith = null;
```

lively.morphic/morph.js
```diff
@@ -270,8 +270,7 @@ export class Morph {
         defaultValue: pt(0, 0),
         get () {
           let layout;
-          if (layout = this._askLayoutForBounds) {
-            this._askLayoutForBounds = false;
+          if (layout = this.renderingState.cssLayoutToMeasureWith) {
```
linusha, line 273: This results in an assignment and afterwards a check if the assigned value was truthy?

lively.morphic/rendering/renderer.js
```diff
@@ -190,9 +187,22 @@ export default class Renderer {
       this.handleAddedAnimationChange(morph);
     }
 
-    for (let morph of this.renderedMorphsToBeMeasured) {
-      const affectedLayout = morph.renderingState.cssLayoutToMeasureWith;
+    for (let [morph, affectedLayout] of this.renderedMorphsToBeMeasured) {
+      // the problem is that this remeasure will in turn lead to a next render
+      // CSS layouts should be able to make conclusive renders in ONE pass
       morph.renderingState.cssLayoutToMeasureWith = false;
```
linusha, line 193:
```suggestion
      morph.renderingState.cssLayoutToMeasureWith = null;
```

merryman (approved):

PR #467, opened 2022-09-08
☁️: remove lively.sync files

merryman:
Since there is not use of this package for the foreseeable future, we remove it from the repo.


linusha (changes requested):
Please remove from the README as well.

linusha (approved):

PR #468, opened 2022-09-08
🪞: remove lively.mirror files

merryman:
Since we are not going to investigate this project further, we remove this subrepo.


linusha (changes requested):
Let's remove the entry from the README as well :)

linusha (approved):

PR #471, opened 2022-09-12
Fix/freezer issues

merryman:
Fixes a syntax errors and issues with the source code transformation in the bundler.


lively.freezer/src/bundler.js
```diff
@@ -617,6 +620,7 @@ export default class LivelyRollup {
 
       const imports = [];
       const toBeReplaced = [];
+
```
linusha, line 623:
```suggestion
```

linusha (approved):

PR #477, opened 2022-09-15
🧰🎨: fix resizing to fit window in world

merryman:
Fixes an issue where the world would not resize exactly to the browser window bounds when the browser window is resized.


linusha (approved):
Works for me! Nice!

PR #479, opened 2022-09-15
🎨: render text selection layer with integer coordinates

merryman:
This may fix the issue that was reported by @linusha on displays that have different zoom factors applied by the OS.


linusha:
Unfortunately, it does not :pensive: 

image

PR #503, opened 2022-10-05
Improved package filtering

### LINKED ISSUE #437: System Browser shows a lot of seemingly duplicate packages
This is open for discussion and I would be interested in any input:

The system browser currently shows a lot of packages. Only the smallest fraction of these are `lively.next` core packages, most are dependencies.

**This is the first thing I am wondering about - can we make this more user friendly? Most people will not want to interact with any of the dependencies. On the other hand, it might be handy to look into dependency code at some times for some special use cases. I am wondering if we can find a nice solution which makes this optional.**

To make things worse, some packages are duplicated in the browser, without any indication of why that is/what the difference is between these!
Take a look at this screenshot that I took from the latest master: 
image

These entries all look the same. Turns out, due to the unlimited power of flatten, we have a bunch of different kind-of versions installed. Thus, these entries are in fact different, but to find that out, one would need to investigate each package.json in all of these packages. Even worse - afterwards one would still have no indication of which version maps to which entry.

**Without getting more conrete, I think the solution here should be simple - we need a nice way to display the version number somehow earlier in the browsing process**.

linusha:
Closes #437.


merryman (changes requested):
Remaining Bugs:
1. Unintended deselect of dependency packages.
2. Typo: `Dependecy`.

merryman (approved):

PR #522, opened 2022-10-14
👼: stop dragging in properties panel from moving panel's target

linusha:


merryman:
From what I understand this is a problem in how the top bar invokes the custom drag.
Concretely in `lively.ide/studio/top-bar.cp.js` it should check for if the event position is *actually inside* the halo when it starts to invoke the custom drag. I think that would be the simpler solution here.

merryman (approved):

PR #524, opened 2022-10-14
🧰🎨: move popups in front of halo

### LINKED ISSUE #347: Mouse events getting picked up by underlying morph
The screenshot is here.  What happens is I bring up a popup (detail design, window in component, etc) and when I click on a title bar or UI element the underlying morph grabs the event.  The world is https://matt.engagelively.com/worlds/load?name=DAN4_2-robin

image

linusha:
Closes #347.

This is more of a proposal than something that I have a hard opinion on. The problem that the issues describes is obvious and I would think this is the canonical fix for it, however, I am not sure if this could have side-effect that I have not thought of that would cause problems.
In case you are not sure if this is problematic, I would probably vote for not merging this and closing the ticket as "won't do". I'd also assume, that the underlying reason for this occurring, opening popups in a rather crowded world, becomes less of a problem once we have the "endless scrolling canvas style world".

### DIFF
--- a/lively.ide/studio/controls/border.cp.js
+++ b/lively.ide/studio/controls/border.cp.js
@@ -1,5 +1,5 @@
 import { Color, rect, Rectangle, pt } from 'lively.graphics';
-import { TilingLayout, Morph, Label, ViewModel, part, add, component } from 'lively.morphic';
+import { TilingLayout, Morph, Label, part, add, component } from 'lively.morphic';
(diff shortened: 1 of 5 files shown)


merryman:
The only problem with this solution is that you wont be able to edit popups themselves via a halo. So wanting to define a new popup via halo will require you to manually remove the style classes before you can proceed to work with it.
I would prefer a solution that augments the styleClasses from the viewModel, so it only has an effect when the popup behavior is activated.
Regarding the z-index solution: I am fine with having a halo class and a popup class with separate z-indexes. But we need to watch this space and make sure this does not grow to more classes. When we have 10 style classes with varying z-indices we know we have a problem.

merryman (approved):

PR #542, opened 2022-10-20
🧰: maximize collapsed browser when opening test

merryman:
When opening a test from the test runner that is already opened in a browser in the world, that browser was previously not uncollapsed if its window was minimized.


lively.ide/test-runner.js
```diff
@@ -308,19 +309,16 @@ export default class TestRunner extends HTMLMorph {
...
   }
 
   async jumpToTest (test, file) {
     const li = LoadingIndicator.open('Scanning source code...');
     try {
       const browser = await this.findBrowserForFile(file);
-      browser.getWindow().activate();
+      const win = browser.getWindow();
+      if (win.minimized) win.minimized = false;
+      win.activate();
```
linusha, line 321: Just a thought - since the minimization/maximization of windows is animated, do we need to await something here?

linusha (approved):

PR #543, opened 2022-10-20
🧰: fix jiggling of number inputs in layout alignment flap

merryman:
Fixes a jiggling effect of the padding inputs on the layout alignment flap.

### DIFF
--- a/lively.ide/studio/controls/layout.cp.js
+++ b/lively.ide/studio/controls/layout.cp.js
@@ -532,9 +532,7 @@ const LayoutControl = component(PropertySection, {
     }]
   }))]
 });
-// AutoLayoutAlignmentFlap.openInWorld()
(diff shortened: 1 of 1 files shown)


linusha (approved):
Just for my understanding - the jiggle occurred due to a "late layout"? Meaning first, the extent out of the definitions was used and later the layout kicked in making the widgets jiggle when the extent was changed? And now we let the layout dictate the extent from the very beginning, eliminating the jiggle?

merryman:
> Just for my understanding - the jiggle occurred due to a "late layout"? Meaning first, the extent out of the definitions was used and later the layout kicked in making the widgets jiggle when the extent was changed? And now we let the layout dictate the extent from the very beginning, eliminating the jiggle?

Exactly right.

PR #544, opened 2022-10-23
Fix/dynamic property reset

merryman:
Fixes accidental resets of values that are caused when we add/remove dynamic properties in the properties panel. E.g. Try to add an opacity effect and then an additional drop shadow effect. The opacity will magically reset to 1, although that was not intended.

### DIFF
--- a/lively.ide/studio/controls/body.cp.js
+++ b/lively.ide/studio/controls/body.cp.js
@@ -1,10 +1,10 @@
-import { TilingLayout, component, ViewModel, part } from 'lively.morphic';
+import { TilingLayout, ShadowObject, component, ViewModel, part } from 'lively.morphic';
 import { Color, rect, pt } from 'lively.graphics';
 import { obj, arr } from 'lively.lang';
 import { once, noUpdate, connect } from 'lively.bindings';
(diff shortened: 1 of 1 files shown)


lively.ide/studio/controls/body.cp.js
```diff
@@ -143,7 +140,7 @@ export class BodyControlModel extends PropertySectionModel {
   activate () {
     this.view.layout = this.view.layout.with({ padding: rect(0, 10, 0, 10) });
     this.view.master = this.activeSectionComponent; // eslint-disable-line no-use-before-define
-    this.addDynamicProperty();
+    this.addDynamicProperty(false, false);
```
linusha, line 143:
```suggestion
    this.addDynamicProperty(null, false);
```
I know that JS does not care (which is a feature by the way :joy: :upside_down_face:), but I like to use `false` for actual boolean values. Here, we want to explicitly pass *nothing*, for which I propose to use `null`. This way, it is also possible to differentiate between explicit and implicit setting a value to nothing (`null` vs `undefined`).

linusha (approved):

Issue #529, opened 2022-10-20
Restructure and clean up the Default Components that come with lively


linusha:
This is quite a large undertaking, but it will pay off since it will allow for faster development of new applications once the new Component Browser (#528) is available.

The structure of some of the components in IDE and and components is very convoluted, e.g., the Popup Windos have a "circular hierarchy" from a dark component to a light one and then another dark one.
This should be cleaned up.

We could also think about migrating the remaining components that are not yet implement with the new component architecture.

linusha:
This includes a actual curation of the popular collection in the components browser. Currently, this is only working on a technical level.

linusha:
One part of this ticket will be a cleanup of the prompts, which are pretty messy with regards to their hierarchy, which leads to unintended behavior in some cases.

Issue #454, opened 2022-09-06
Remove all references to old parts


linusha:
There are still some remining references to the old master components as stored in snapshots loaded via `part://`. Those should be removed, as well as the remaining snapshots.

The code for loading these is already gone, but removing all references should go with an analysis of what kind of tools are impacted and how they are to be replaced within the new project management approach.

PR #547, opened 2022-10-27
Removal of Horizontal and Vertical Layouts

linusha:
As discussed, I went ahead and removed Vertical- and Horizontal Layouts. There are still some references remaining. However, these are in parts of the System that are currently not used as far as I can see or will be refactored in our ongoing project anyways. See the following list:

- Responsive Halo -> will be fixed as part of the current project
- Component Browser -> will be fixed as part of the current project
- MorphicDBWidget and related components -> **I have no idea what this is or how it is used?**
- Reconciliation Tests -> @merryman can you take care of this once we have merged this?
- Text UI -> will be fixed as part of the current project
- partsbin -> to be deleted (see #454)
- Version Control -> **I have no idea what this is or how it is used?**
- Layout Halos -> will be fixed as part of the current project, **I think?**. Otherwise, I'd propose removing what we do not need anymore.
- `components/widgets.js` -> as far as I can see, `LabeledCheckBox` and `DropDownSelector` are mostly duplicates of widgets that are duplicated in other places (with `ViewModels`) and are mainly used in the Layout Halos. No harm here, as far as I see it. However, stuff like this should probably be cleaned up (e.g., during #529)

@merryman would you mind checking if we agree on all points in the list? We can also go over it in a quick call whenever the opportunity arises, as this is not urgent.

---

While at it, I also refactored our tooltips to utilize the `ViewModel` architecture. I intended for `part(SystemTooltip)` to be the new canonical way of opening tooltips. However, something very weird is happening - when I open a new world, hovering will not display any tooltips and causing a tooltip by e.g., creating a new morph with the topbar will throw errors. **Opening the tooltips.cp.js file in the SystemBrowser and saving it will solve this.** I find this to be strange, as the module is loaded without it being opened in the SystemBrowser, so this is not the cause. @merryman do you have any idea what is going on here? :O


merryman:
**1. MorphicDBWidget and related**
These are tools that allow for inspection and managing of the object db that stores parts and worlds.
These tools used to be handy/necessary because back in the day we had everything revolving around snapshots that could either be stored locally on the lively instance on the machine (implemented via a sqlite backend upon which we run pouchdb) or versioned to a remotely accessible lively.next server that has a fully fledged couchdb instance running which is able to perform the merging and reconciliation of commits. These tools would allow to directly manage the state of any of these databases by connecting to them (locally or remotely) or allow to push and pull data from or to a local/remote database. As long as we do not have a sound snapshot development process in place it will be not really useful to revisit these tools for now. But it is something to keep in mind.

**2. Regardin the tooltips**
This is probably a circular dependency. I will investigate.

merryman (approved):

linusha:
> But I could not find any changes related to that in your PR. Can you point to the files you are referring to?

Just so that we are on the same page: The files/tools mentioned in the above list are all untouched, at least in this PR. The reason I listed them was to make sure that we both either think a) we will fix them in the near future anyways or b) they are not used currently/in the near future. My goal was to not have stuff that is used broken by referencing the old layouts, that are now gone.
In the case that we need to touch tools (like the Components Browser) anyways, there are now broken Layout references, but those can be fixed up then.

The Version Control tool I referred to is `version-control.js` in `lively.ide`

@merryman I will refrain from merging this until you report back with some insights regarding the tooltips, ok? :slightly_smiling_face:

merryman:
> @merryman I will refrain from merging this until you report back with some insights regarding the tooltips, ok? 🙂

The problem with the tooltips is that you import `Label` and `ViewModel` from the `lively.morphic/index.js`. Since the `index.js` imports a vast set of files in the system, this is a reliable way to generate circular dependencies. It is advised to basically never import the `index.js` from within of `lively.morphic` but instead directly import the module the entity is declared in.

merryman:
Aside from this there are still a couple of layout issues in the branch. Here is 2 I found:

When the loading indicator displays a longer label it looks like the layout is wrapping the morphs instead of adjusting the width of the loading indicator:
broken loading indicator

The version checker looks like this for me:
broken version checker

lively.morphic/tooltips.cp.js
```diff
@@ -1,28 +1,61 @@
 import { component } from './components/core.js';
 import { pt, Color } from 'lively.graphics';
-import { HorizontalLayout, Label } from 'lively.morphic';
+import { TilingLayout } from './layout.js';
+import { Label, ViewModel } from 'lively.morphic';
```
merryman, line 4: Do not import from the `index.js`. Instead import from the modules these entities are declared in directly.

linusha: Perfect, that works! Thanks!

linusha:
> When the loading indicator displays a longer label it looks like the layout is wrapping the morphs instead of adjusting the width of the loading indicator: broken loading indicator

This should be fixed now. It even works for status messages that contain line breaks.

image

I took care of the version picker as well.

@merryman any more Issues you found? Otherwise, feel free to push the button :grinning:

merryman:
> > When the loading indicator displays a longer label it looks like the layout is wrapping the morphs instead of adjusting the width of the loading indicator: broken loading indicator
> 
> This should be fixed now. It even works for status messages that contain line breaks.
> 
> image
> 
> I took care of the version picker as well.
> 
> @merryman any more Issues you found? Otherwise, feel free to push the button 😀

I found that the components browser still references `VerticalLayout`.

linusha:
> I found that the components browser still references `VerticalLayout`.

Yes. As I wrote in my comment above, or to quote the original PR description:

> There are still some references remaining. However, these are in parts of the System that are currently not used as far as I can see or will be refactored in our ongoing project anyways.

I can of course go ahead and just plain replace the layouts with a `new TilingLayout()` call in all mentioned tools, but I think it makes more sense to keep the layout definition as it was intended, to actually fix it later on, no?
I simply had no desire to guess what layout configuration could be correct in tools that are either broken or unused at the moment. I think it makes more sense to tackle this when the tools itself get fixed.

PR #550, opened 2022-10-31
Object Equals for circular objects

### LINKED ISSUE #549: obj.equals ends up in infinite loop when comparing circular objects.
Encountered by @linusha when implementing the new renderer (sudden freeze of the lively session), `obj.equals` loops indefinitely when comparing two objects that contain circular references.

Solution to this would be to implement an efficient strategy that allows us to stop before we recursively traverse an object that we have previously already visited.

linusha:
Closes #549.

Unfortunately, the linter came with a Klappstuhl in array.js. All those changes are in a separate commit however, so that it should be alright if you take a look at the changes commit-wise and not all at once :)

### DIFF
--- a/lively.lang/array.js
+++ b/lively.lang/array.js
@@ -370,13 +370,14 @@ function equals (array, otherArray) {
  * (`lang.obj.equals`) to its corresponding element in `otherArray`.
  * @param {any[]} array
  * @param {any[]} otherArray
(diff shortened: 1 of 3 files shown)


merryman (approved):
I guess all we can do here is to trust Chai.JS folks on their implementation of equals here. But I also dont see any better alternatives.

PR #551, opened 2022-11-04
Text menu

linusha:
This solves multiple problems related to the editing of rich text in `Text` morphs and also provides an interface for all text formatting options that we support, that were previously partially unavailable via GUI.

This required quite a lot of changes, most of which can be seen in this demo video:

https://user-images.githubusercontent.com/14252419/203550006-4acbc91f-80ed-4cae-9323-14d68209e9e3.mp4

Here is a hopefully comprehensive list of changes contained in this PR:

1. Creating a text with the topbar will now focus it for editing.
2. This editing mode will be left, when the text morph looses focus.
3. This editing mode can be reentered by clicking double-clicking the border of the text's halo.
4. The problems with selections and halos (#560) are fixed.
5. Selecting text in editing mode will open a text formatting popup with which the rich text formatting of the selected text can be changed. The side bar will always change the formatting of the whole morph. There exists a flag that enables/disables this behavior. For convenience, I introduced a `Text` subclass. The described selection based formatting does not support multi-select.
6. The side bar provides a widget to change the padding of a text.
7. The side bar provides means of changing the embedding of a morph in text when it is selected.

One problem still exists, but I believe this is related to layouts and I propose proceeding with this anyways. The `PaddingControls` "jump" around when toggling extended view. This can be seen at the end of the above video.


merryman:
It works very well in my testing so far. Very nice!
I have one question and 2 critique points:

1. Is there a particular reason why we can only enter the edit mode by double clicking the halo border? I feel awfully clumsy to me, but maybe I don't see the problems with turning a plain double click on the halo to enter the edit mode as well.
2. I am one of these crazy people that like to select by double clicking a word. In those cases the popup does not seem to appear yet which is inconvenient.
3. Question: There are a couple of design glitches that I guess are much easier to fix by myself than carrying out in a discussion here. If you don't mind, I can just add some commits to this branch after we are done with the 2 things above.

linusha:
Alright, should be done.

Allowing to enter the edit-mode with a double click made it necessary to change when a Halo appears/disappears, as previously, a click on an active Halo would disable that Halo. I believe this change does not lower the usability of the system, just pointing it out.

merryman:
I found two more usability issues:

1. The popup seems to disappear int between successively selecting words via double click:
popup-disappearing
2. The popup sometimes overlaps the cursor and the text it is modifying. I dont know whats a good solution to that. Memorizing the last position the popup has been dragged to explicitly? Or some smart logic that factors in the text bounds of the text morph?
popup-overlapping

lively.ide/text/rich-text-commands.js
```diff
@@ -156,6 +156,7 @@ export var commands = [
     name: 'open text attribute controls',
     exec: async text => {
       const { module } = await System.import('lively.modules');
+      // FIXME: use new formatting panel here
```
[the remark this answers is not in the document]
linusha: Thinking about it, I removed the command completely. Opening the popup (i.e., selection based formatting options) does not really make sense via the keyboard from a UX standpoint in my opinion, as the popup can only be used via mouse.
Opening the properties panel (i.e., the global text formatting options) has nothing to do with rich-text-commands per se.

We could think about providing keyboard shortcuts for rich text formatting (e.g., Ctrl+B -> bold text), but I think this is a separate feature and not urgent.

lively.ide/js/inspector/index.js
```diff
@@ -388,7 +388,7 @@ export class PropertyControl extends DraggableTreeLabel {
       inspector.openWidget = editor;
       editor.viewModel.startPadding(target[keyString]);
       await editor.fadeIntoWorld(evt.positionIn(target.world()));
-      connect(editor.viewModel, 'paddingChanged', (padding) => {
+      connect(editor.viewModel.ui.paddingControls.viewModel, 'paddingChanged', (padding) => {
```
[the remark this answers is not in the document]
linusha: Done.

lively.ide/studio/controls/popups.cp.js
```diff
@@ -320,20 +316,58 @@ export class PaddingPopupModel extends PopupModel {
...
+            { target: 'close button', signal: 'onMouseDown', handler: 'close' }
+          ];
+        }
+      }
+    };
+  }
+
+  viewDidLoad () {
+    connect(this.ui.paddingControls.viewModel, 'paddingChanged', this, 'paddingChanged');
+  }
```
merryman, line 356: replace this one with a binding instead please.

lively.ide/studio/controls/popups.cp.js
```diff
@@ -745,119 +779,170 @@ const InsetShadowPopup = component(ShadowPopup, {
...
+      textAndAttributes: ['', {
+        fontSize: 18,
+        textStyleClasses: ['material-icons']
+      }]
+    })
+  ]
+}
+);
+
+export const PaddingControlsDark = component(PaddingControlsLight, {
```
merryman, line 887: I am not entirely happy with the look of the padding controls yet. I would like this to basically look more like the border radius control in the shape section of the side bar. 

For recap: 
border radius control

lively.morphic/text/morph.js
```diff
@@ -2430,16 +2430,24 @@ export class Text extends Morph {
 
     if (evt.isShiftDown()) {
       this.selection.lead = clickTextPos;
+      if (this.editorModeName === 'richText') this.editorPlugin.showFormattingPopUp();
```
merryman, line 2433: A little more brief, assuming no other plugin implements this method:
`this.pluginInvokeFirst('showFormattingPopUp')`

merryman (changes requested):

lively.ide/studio/controls/popups.cp.js
linusha: Now it looks like this:

textcontrols

merryman: It's beautiful.

PR #552, opened 2022-11-04
🎨: do not fail when binding target is removed from component

linusha:
This is an interesting problem that I ran into. I am very positive that we need to find a nice solution to the underlying problem - if this solution is, it I am not so sure about. Thus, this is a RFC as well as a PR :D

**What is the underlying problem to solve**: 

Generally, one great thing about component systems where view and behavior are separated (besides all the non-coder boogaloo :slightly_smiling_face:) is that they allow for great reuse of code, especially on the model side, i.e., I can build several different front ends that serve similar purposes in different contexts and just attach the same model.

Concretely, I stumbled upon this problem when building a kind of "rich text formatting halo", that should provide a subset of the functionality of the text controls currently contained in the sidebar. Therefore, all functionality is already implemented in the text control model, so I just threw together the UI elements, and as long as one uses the same names as in the original component, one can reuse the model, which is great. However, removing/not including an element which is part of bindings in the `ViewModel`, will lead to an error when the bindings get reified.

I think there should be the possibility to not have this error occurring.

### DIFF
--- a/lively.morphic/components/core.js
+++ b/lively.morphic/components/core.js
@@ -358,7 +358,7 @@ export class ViewModel {
           updater
         });
       } catch (err) {
-        console.warn('Failed to reify biniding: ', target, model, signal, handler);
+        if (System.debug) console.warn('Failed to reify binding: ', target, model, signal, handler);
       }
     }
   }


merryman:
I am actually surprised that this was not already supported. I have always considered this to be nice feature of the view model architecture, where you can also attach a model to a less capable UI that supports less inputs yet displays all the information required.

lively.morphic/components/core.js
```diff
@@ -344,7 +344,8 @@ export class ViewModel {
       override = false, converter = false, updater = false
     } of this.bindings) {
       try {
-        if (model) target = this.view.getSubmorphNamed(model).viewModel;
+        // safe navigation allows reuse of view models for UIs with a subset of functionality
+        if (model) target = this.view.getSubmorphNamed(model)?.viewModel;
```
merryman, line 348: I dont know, maybe we can still place a `console.warn` here? Just in case someone does this by accident, or has the name of the input mixed up.

linusha: I used sloppy language in the original description, sorry. This is actually working as it is now, i.e., I can build "sub-set" components with the "original" `ViewModel`.
What happens is that the line that I changed here fails, which takes us into the `catch` block, leading to a `console.warn`. So we probably have exactly what you wish for? My change only leads to us not entering the `catch` block in the case of a "sub component`".

I see why such a warning can be of help, but found them to become quite spamy. The reason is, that my component is a kind of popup, that gets reinstantiated quite often.

However, I absolutely see why we would keep the warning around, your call. Feel free to close this!

merryman: I would be open to use the `System.debug` flag in order to limit the warnings we send out for view models. We already do that for quite a bit of warning logs in the system.

merryman (approved):

PR #553, opened 2022-11-07
Fix inspector widgets

linusha:
This resolves some minor UX problems that occured presumably since we switched to the new PopUp Widgets.

Namely, now:

- only one widget can be open at any time,
- scrolling or switching the inspectors target will close any open widgets,
- closing the window of the inspector will close any open widgets.

### DIFF
--- a/lively.ide/js/inspector/index.js
+++ b/lively.ide/js/inspector/index.js
@@ -377,10 +377,12 @@ export class PropertyControl extends DraggableTreeLabel {
   }
 
   static renderRectangleControl (args) {
-    const { value, keyString, valueString, target, node } = args;
+    const { keyString, valueString, target, node, tree } = args;
+    const inspector = tree.owner.viewModel;
     const handler = async (evt) => {
       // fixme: add rectangle popup!
       const editor = part(PaddingPopup, { hasFixedPosition: true });
+      inspector.openWidget = editor;
       editor.viewModel.startPadding(target[keyString]);
       await editor.fadeIntoWorld(evt.positionIn(target.world()));
       connect(editor.viewModel, 'paddingChanged', (padding) => {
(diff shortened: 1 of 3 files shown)


lively.ide/js/inspector/index.js
```diff
@@ -356,6 +355,7 @@ export class PropertyControl extends DraggableTreeLabel {
           node.rerender();
         }
       })));
+      tree.owner.openWidget = menu;
```
merryman, line 358: Doesnt that need to be the view model?

lively.ide/js/inspector/index.js
```diff
@@ -492,25 +500,27 @@ export class PropertyControl extends DraggableTreeLabel {
...
 
   static renderColorControl (args) {
-    let propertyControl; const {
-      node, gradientEnabled, fastRender,
-      valueString, keyString, value, target
+    const {
+      node, gradientEnabled,
+      valueString, keyString, value, target, tree
     } = args;
+    const inspector = tree.owner;
```
merryman, line 518: I think we want the view model here too.

merryman: I see that this works by exposing the viewModel property, but still. I would want to have a consistent access.

linusha: Good catch. Thanks!

linusha:
I unified the access, but the other way around actually :see_no_evil:.
If I remember correctly, we discussed this some time before and were quite sure that typing `.viewModel` is a code smell in most cases. Do you think otherwise now, or is this fine with you?

linusha:
@merryman would you approve this, then? :slightly_smiling_face:

merryman (approved):

PR #554, opened 2022-11-08
🧰: add debug console

merryman:
Carries over the previous parts based implementation of Console into a component based definition.

### DIFF
--- a/lively.ide/debug/console.cp.js
+++ b/lively.ide/debug/console.cp.js
@@ -1,0 +1,263 @@
+import { ViewModel, part, Text, component } from 'lively.morphic';
+import { obj, string } from 'lively.lang/index.js';
+import { pt, Color } from 'lively.graphics';
+
+const defaultConsoleMethods = ['log', 'group', 'groupEnd', 'warn', 'assert', 'error'];
+
+function formatTemplateString (template, ...args) {
+  let string = template;
+  for (let i = 0; i < args.length; i++) {
+    let idx = string.indexOf('%s');
+    if (idx > -1) string = string.slice(0, idx) + String(args[i]) + string.slice(idx + 2);
+    else string = string + ' ' + String(args[i]);
+  }
+  return string;
+}
+
+function installErrorCapture (target, _window = window) {
+  if (!target._errorHandler) {
+    target._errorHandler = (function errorHandler (errEvent, url, lineNumber, column, errorObj) {
+      let err = errEvent.error || errEvent; let msg;
+      if (err.stack) msg = String(err.stack);
+      else if (url) msg = `${err} ${url}:${lineNumber}`;
+      else msg = String(err);
+      if (typeof target.error === 'function') target.error(msg);
+    });
+    _window.addEventListener('error', target._errorHandler);
+  }
+  if (!target._errorHandlerForPromises) {
+    target._errorHandlerForPromises = function unhandledPromiseRejection (evt) {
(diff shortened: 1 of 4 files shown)


linusha (changes requested):
1. This is a cool tool!
2. I propose we should bring at least the world command to open this tool up back, would you agree?
3. I would like to have comments here, as the workings of this tool are not super straight forward. If that is okay with you, I'd like to add them myself and have you review them, to make sure I understood correctly what is going on here. Just leave me a comment if that works for you...
4. To me this seems to be only a "half way" usage of the `ViewModel` architecture? I think I would like to get rid of the `LocalJSConsole` subclass of `Text` and move the functionality into the `ViewModel`. Or am I missing something? 
I think this also causes problems functionality wise as it is, as I am not able to invoke the static `open` method, as thus unable to open the Console in a window?
If you agree and give green ligh, I will take care of this together with the comments, @merryman.

merryman:
> 1. This is a cool tool!
> 2. I propose we should bring at least the world command to open this tool up back, would you agree?
> 3. I would like to have comments here, as the workings of this tool are not super straight forward. If that is okay with you, I'd like to add them myself and have you review them, to make sure I understood correctly what is going on here. Just leave me a comment if that works for you...
> 4. To me this seems to be only a "half way" usage of the `ViewModel` architecture? I think I would like to get rid of the `LocalJSConsole` subclass of `Text` and move the functionality into the `ViewModel`. Or am I missing something?
>    I think this also causes problems functionality wise as it is, as I am not able to invoke the static `open` method, as thus unable to open the Console in a window?
>    If you agree and give green ligh, I will take care of this together with the comments, @merryman.

Agree to all of the above. Maybe should have turned this into a draft anyway. Feel free to add in your suggestions!

linusha:
> Agree to all of the above. Maybe should have turned this into a draft anyway. Feel free to add in your suggestions!

Done :slightly_smiling_face: 

> That is correct but I ran into issues with the current implementation of view models which somewhat does not work well when we want to extend properties that are defined in the view already. In this case we want to extend the keybindings and commands property, which is already defined on the text itself. However there is no convenient way right now to easily extend those from the view model. We always end up loosing the what is defined on the text morph already...

I see, that makes sense. I experimented a bit with this and have implemented an attempt with `withoutBindingsDo()` which I thought should work (if it would be a nice solution is another matter) but it doesn't. I feel like I have a conceptual misunderstanding there of how some things with the `bindings` work, probably something for a call as well :telephone_receiver:

merryman:
> > Agree to all of the above. Maybe should have turned this into a draft anyway. Feel free to add in your suggestions!
> 
> Done 🙂
> 
> > That is correct but I ran into issues with the current implementation of view models which somewhat does not work well when we want to extend properties that are defined in the view already. In this case we want to extend the keybindings and commands property, which is already defined on the text itself. However there is no convenient way right now to easily extend those from the view model. We always end up loosing the what is defined on the text morph already...
> 
> I see, that makes sense. I experimented a bit with this and have implemented an attempt with `withoutBindingsDo()` which I thought should work (if it would be a nice solution is another matter) but it doesn't. I feel like I have a conceptual misunderstanding there of how some things with the `bindings` work, probably something for a call as well 📞
> 
> @merryman If you are fine with my changes just leave me a comment or something, I will then approve and merge.

Yes let's talk about it. but overall all of this looks decent to me for the time being.

linusha (approved):

PR #556, opened 2022-11-08
🧰: do not show halo for world

merryman:
Small adjustment to the #519. I dont think we want a halo on the world. It almost always obstructs/confuses the user.

### DIFF
--- a/lively.ide/studio/top-bar.cp.js
+++ b/lively.ide/studio/top-bar.cp.js
@@ -529,7 +529,9 @@ export class TopBarModel extends ViewModel {
     if (obj.isArray(targets)) {
       if (targets.length === 0) return;
       halo = this.world().showHaloForSelection(targets);
-    } else halo = this.world().showHaloFor(targets);
+    } else {
+      halo = this.world().showHaloFor(targets);
+    }
     once(halo, 'remove', () => {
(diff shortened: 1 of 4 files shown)


linusha (commented):
This slightly confuses me. Doesn't this change only stops the Halo from being shown on the world when a) we have a ColorPicker whose target is the world and b) we switch that pickers mode from Gradient to Solid? As the only way I managed to open a ColorPicker whose target was world was via the Inspector -> fill property, this seems to be quite an edge case?

What am I missing here? 

---

In the case the actual idea was to never allow Halos on Worlds, I am open to that.

merryman:
> This slightly confuses me. Doesn't this change only stops the Halo from being shown on the world when a) we have a ColorPicker whose target is the world and b) we switch that pickers mode from Gradient to Solid? As the only way I managed to open a ColorPicker whose target was world was via the Inspector -> fill property, this seems to be quite an edge case?
> 
> What am I missing here?
> 
> In the case the actual idea was to never allow Halos on Worlds, I am open to that.

Yes, I agree that this is somewhat of an edge case. I think the suggestion of just disabling halos on the world by default is the best solution.

linusha (approved):

PR #564, opened 2022-11-18
📦: bump required nodejs version in install.sh + start.sh

merryman:
Should capture the issue that was *again* encountered in #561.

@linusha We may need to update the GitHub Actions stuff as well inside this PR.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -13,7 +13,7 @@ jobs:
     - name: Setup node
       uses: actions/setup-node@v2
       with:
-        node-version: '16.13.1'
+        node-version: '18.12.1'
     - name: Install jsdocs
       run: npm install -g jsdoc@3.6.7
     - name: Install jsodc theme
--- a/.github/workflows/check_builds.yml
+++ b/.github/workflows/check_builds.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v2
         with:
-          node-version: '17.8.0'
+          node-version: '18.12.1'
       - name: Install `sultan`
         run: pip3 install sultan 
       - name: Checkout repository
--- a/.github/workflows/check_docs.yml
+++ b/.github/workflows/check_docs.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Setup node
       uses: actions/setup-node@v2
       with:
-        node-version: '16.13.1'
+        node-version: '18.12.1'
     - name: Install jsdocs
       run: npm install -g jsdoc@3.6.7
     - name: Install jsodc theme
--- a/.github/workflows/check_world_loading.yml
+++ b/.github/workflows/check_world_loading.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v2
         with:
-          node-version: '17.8.0'
+          node-version: '18.12.1'
       - name: Checkout repository
         uses: actions/checkout@v3
       - name: Install puppeteer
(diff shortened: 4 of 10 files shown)


linusha:
I find the whole situation to be a bit confusing, to be honest.

I am quite sure that I was able to reproduce the situation Rick had as well, with an error occurring when running `start.sh`, but with `install.sh` succeeding. However, I did not take note of which node version I used to reproduce this. Just running `nvm use 17` gives me `17.9.1` and no problems.

Note, that the actions work as intended as well, and they run build processes (even running `install.sh` runs build processes nowadays!) with node version `17.8.x` for the most part.
Also, the offending line in the traces Rick provided (and they are consistent with what I observed) is line 16 in 52ef511157e536cbf5dfa6075654a977015d45f9 (the safe navigation `?.` syntax). That has been there for months. Why does it fail now? 

This is all quite curious, but in the end probably nothing we can reconstruct. I just wanted to point that out, so that we know what to look out for in the future, if something similar would happen again. It would be beneficial to understand what went wrong here, in order to add a CI check against this problem.

I just went ahead and specified the latest LTS version of `node` in the README, and will change all actions to use that as well.

Feel free to merge after reading this :smile:.

linusha (approved):

merryman:
> I find the whole situation to be a bit confusing, to be honest.
> 
> 
> 
> I am quite sure that I was able to reproduce the situation Rick had as well, with an error occurring when running `start.sh`, but with `install.sh` succeeding. However, I did not take note of which node version I used to reproduce this. Just running `nvm use 17` gives me `17.9.1` and no problems.
> 
> 
> 
> Note, that the actions work as intended as well, and they run build processes (even running `install.sh` runs build processes nowadays!) with node version `17.8.x` for the most part.
> 
> Also, the offending line in the traces Rick provided (and they are consistent with what I observed) is line 16 in 52ef511157e536cbf5dfa6075654a977015d45f9 (the safe navigation `?.` syntax). That has been there for months. Why does it fail now? 
> 
> 
> 
> This is all quite curious, but in the end probably nothing we can reconstruct. I just wanted to point that out, so that we know what to look out for in the future, if something similar would happen again. It would be beneficial to understand what went wrong here, in order to add a CI check against this problem.
> 
> 
> 
> I just went ahead and specified the latest LTS version of `node` in the README, and will change all actions to use that as well.
> 
> 
> 
> Feel free to merge after reading this :smile:.

If that is the case we should not merge this. My understanding was this was solely due to a feature that node 18 supports and we implicitly now assume.
We should review in more detail what's really going on.

merryman:
From the conversation in #561 it appears this has less to do with the nodejs version after all. I will close this for now.

linusha:
> From the conversation in #561 it appears this has less to do with the nodejs version after all. I will close this for now.

Even though #561 is not the cause, I still think we should proceed with this. My two cents:

`node` 17 will not be a LTS version, sticking with that is thus at least bad practice. I am quite sure that we cannot go back to 16, the reason being something with esm support I believe? In the case it helps you remembering, here are some relevant commits:
image
 
This leaves us with 18 being the next LTS version we can use. Ironically, you were the first using 18, which is why we have `--no-experimental-fetch` in `install.sh`. Once we are on a LTS version again, we should begin to better document our version bumps. If nothing comes up that requires a newer version, LTS versions receive patches for 30 months.

What we could think about is if we introduce `--no-experimental-fetch` consistently before merging this, and/or looking into retiring `node-fetch` from `flatn`, which the implementation that now comes with node per default is based on.

merryman:
Added an additional nodejs version check into the `install.sh` itself. That would have prevented rick form running into the issue he previously reported in #561.

PR #577, opened 2022-12-09
🧰: adjust default extent of console

merryman:
This is a small correction to #554 which adjusts the default extent of the console and removes a debugger statement.


merryman:
This still suffers from a bug in the master components. Will add a fix soon.

linusha (approved):

PR #578, opened 2022-12-09
🎛️L add more left padding to loading indicator

merryman:
Adjusts the lefthand padding of the loading indicator to cure us all from the horrible eyesore we were left with.


merryman:
Just realized you already covered that in #567 @linusha.

PR #606, opened 2023-01-06
🧰👤: decouple flap component from sidebars

linusha:
Allows to easily plug in other stuff than the sidebar into the flap components.


lively.ide/studio/sidebar-flap.js
```diff
@@ -0,0 +1,78 @@
...
+export function relayoutSidebarFlapInWorld () {
+  if (this.target === 'scene graph') return;
+  const propertiesPanelExists = $world.get(this.target);
+  if (propertiesPanelExists) this.view.position = this.view.position.withX($world.visibleBounds().width - defaultPropertiesPanelWidth - this.view.width);
+  else this.view.position = this.view.position.withX($world.visibleBounds().width - this.view.width);
+}
+
+export async function openSidebarFlapInWorld () {
+  const world = this.world();
+  const { view } = this;
```
merryman, line 49: I gotta say, I would prefer this to be passed as an argument. I know this is technically possible but it just is super confusing to me to assume that when we are executing we are bound against a flap model.

lively.ide/studio/flap.cp.js
```diff
@@ -0,0 +1,77 @@
...
+  async executeAction () {
+    this.action.bind(this)();
+  }
+
+  onWorldResize () {
+    this.relayoutRoutine.bind(this)();
+  }
+
+  openInWorld () {
+    this.openingRoutine.bind(this)();
```
merryman, line 43: Can we just use these as plain callbacks without binding them dynamically agains the view model? We can just pass the view model as an argument to the callbacks. That way the callbacks being defined outside of the view model class could contain a parameter name like `flapViewModel` which makes things more clear than implicitly assuming this refers to the flap view model.

merryman (changes requested):

linusha:
Superseded by #613.

PR #611, opened 2023-01-09
🧰: avoid empty inspector when printing of values fails

### LINKED ISSUE #610: Blank inspector on some imported libraries
To reproduce, in a JavaScript workspace, execute:
```
import { Chart, Legend, ArcElement, PieController, registry, registerables } from 'esm://cache/chart.js@3.9.1';
registerables
```
then cmd-shift-I on `registerables`
You'll see:
inspector-bug

The inspector window shows nothing and can't be closed.

merryman:
Fixes empty inspectors when inspecting certain data structures.
It happens when some of the values inside a structure that are supposed to be displayed in the inspector fail to convert to a string. I replaced the printing approach in the inspector with a `safePrint` that attempts different stringification attempts and wraps each in a try/catch.

closes #610.

### DIFF
--- a/lively.ide/js/inspector/context.js
+++ b/lively.ide/js/inspector/context.js
@@ -107,7 +107,7 @@ function safePrint (value) {
 
   }
 
-  if (!result) return '[CAN NOT BE DISPLAYED]';
+  if (!result) return '[CANNOT BE DISPLAYED]';
   return result;
 }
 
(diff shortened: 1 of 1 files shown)


lively.ide/js/inspector/context.js
```diff
@@ -92,17 +92,37 @@ export function isMultiValue (foldableValue, propNames) {
...
+
+  try {
+    if (!result) {
+      result = JSON.stringify(value);
+    }
+  } catch (err) {
+
+  }
+
+  if (!result) return '[CAN NOT BE DISPLAYED]';
```
linusha, line 110:
```suggestion
  if (!result) return '[CANNOT BE DISPLAYED]';
```

linusha (approved):
I would appreciate the typo being fixed, otherwise :+1:

PR #662, opened 2023-01-26
🧰: prevent inconsistent properties control state with opened popups

### LINKED ISSUE #658: Should Popups block their "parent" controls in property panel?
color-picker-inconsistent-state

As can be seen, it is currently possible to create inconsistent state with the color picker opened. This might very well apply to other widgets/properties as well.
I see two options to tackle this:

1. We block the color editor in the properties panel as long as the color picker is open.
2. We hook them up bidirectional (changing the value in the color picker will update the properties panel).

linusha:
Closes #658.

### DIFF
--- a/lively.ide/studio/controls/border.cp.js
+++ b/lively.ide/studio/controls/border.cp.js
@@ -98,11 +98,20 @@ export class BorderControlModel extends PropertySectionModel {
               target: 'more button',
               signal: 'onMouseDown',
               handler: 'openPopup'
+            },
+            {
+              signal: 'onMouseDown',
+              handler: 'onMouseDown'
             }
           ];
         }
       }
     };
+  }
+
+  onMouseDown () {
+    this.closePopup();
+    this.models.borderColorInput.closeColorPicker();
   }
 
   onRefresh (prop) {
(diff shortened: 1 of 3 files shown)


lively.ide/studio/controls/border.cp.js
```diff
@@ -75,6 +75,12 @@ export class BorderControlModel extends PropertySectionModel {
         }
       },
 
+      expose: {
+        get () {
+          return ['onMouseDown'];
+        }
+      },
+
```
merryman, line 83: Better to work with a binding here, since exposing will override the default onMouseDown of this view entirely (without calling super).

lively.ide/studio/properties-panel.cp.js
```diff
@@ -234,6 +234,11 @@ export class PropertiesPanelModel extends ViewModel {
 class BackgroundControlModel extends ViewModel {
   static get properties () {
     return {
+      expose: {
+        get () {
+          return ['onMouseDown'];
+        }
+      },
```
merryman, line 241: dito

lively.ide/studio/controls/fill.cp.js
```diff
@@ -17,6 +17,11 @@ export class FillControlModel extends ViewModel {
       placeholderImage: {
         defaultValue: placeholderImage
       },
+      expose: {
+        get () {
+          return ['onMouseDown'];
+        }
+      },
```
merryman, line 24: dito

merryman (changes requested):

merryman (approved):

PR #663, opened 2023-01-27
Enhancement/direct icon insertion

linusha:
icon_button

I can type super well!

### DIFF
--- a/lively.halos/morph.js
+++ b/lively.halos/morph.js
@@ -2008,7 +2008,7 @@ export default class Halo extends Morph {
     const t = this.target;
     const topBar = this.topBar;
 
-    if (!t.isText) return;
+    if (!t.isText && !t.editorModeName === 'richText') return;
 
     // this makes sense even if target is not readonly
     // in the case we are in halo mode, this allows for editing which would be otherwise blocked by the halo
@@ -2022,9 +2022,9 @@ export default class Halo extends Morph {
     });
 
     connect($world, 'onMouseDown', t, 'cancelTemporaryEdit');
-    t.showIconButton();
     // switch to hand mode to stop halo from eating clicks for editing
     topBar.setEditMode('Hand', true, true);
+    t.editorPlugin.showIconButton(true);
     this.remove();
   }
 
--- a/lively.halos/morph.js
+++ b/lively.halos/morph.js
@@ -2022,7 +2022,7 @@ export default class Halo extends Morph {
     });
 
     connect($world, 'onMouseDown', t, 'cancelTemporaryEdit');
-
+    t.showIconButton();
     // switch to hand mode to stop halo from eating clicks for editing
     topBar.setEditMode('Hand', true, true);
     this.remove();
--- a/lively.halos/morph.js
+++ b/lively.halos/morph.js
@@ -2004,30 +2004,6 @@ export default class Halo extends Morph {
   // ui events
   // -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
 
-  temporaryEditTextMorph (evt) {
-    const t = this.target;
-    const topBar = this.topBar;
-
-    if (!t.isText && !t.editorModeName === 'richText') return;
-
-    // this makes sense even if target is not readonly
-    // in the case we are in halo mode, this allows for editing which would be otherwise blocked by the halo
-    t.prevReadOnly = t.readOnly;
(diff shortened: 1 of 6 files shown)


lively.morphic/text/morph.js
```diff
@@ -3933,11 +3935,40 @@ export class Text extends Morph {
...
+          submorphs: [
+            iconButton]
+        });
+        iconButton.master = { auto: PropertyLabel, hover: PropertyLabelHovered };
+        this.iconButton = iconButton;
+        this.iconButton.onMouseDown = () => this.execCommand('add icon at cursor position');
+        iconButtonHolder.openInWorld(this.topRight);
+      });
+  }
+
```
merryman, line 3963: How about we put this logic into a command that we move into the editor plugin? That way we can get rid of this weird async import stuff.

lively.ide/text/rich-text-editor-plugin.js
```diff
@@ -34,6 +35,32 @@ export class RichTextPlugin extends EditorPlugin {
...
+      layout: new TilingLayout({
+        hugContentsVertically: true,
+        hugContentsHorizontally: true
+      }),
+      submorphs: [
+        iconButton]
+    });
+    iconButton.master = { auto: PropertyLabel, hover: PropertyLabelHovered };
+    this.textMorph.iconButton = iconButton;
+    this.textMorph.iconButton.onMouseDown = () => this.textMorph.execCommand('add icon at cursor position');
```
merryman, line 56: Can we use a `connect()` binding on `onMouseDown` instead of monkey patching it here? The current solution will for instance block the default behaviour of `onMouseDown` which is, among other things, the application of master components etc.

merryman, line 56:
```suggestion
    connect(this.textMorph.iconButton, 'onMouseDown', () => this.textMorph.execCommand('add icon at cursor position'));
```

lively.morphic/text/morph.js
```diff
@@ -3986,9 +3959,10 @@ export class Text extends Morph {
     topBar.setEditMode(topBar.recoverMode, true);
     this.readOnly = this.prevReadOnly;
     this.collapseSelection();
-    this.editorPlugin?.removeFormattingPopUp && this.editorPlugin.removeFormattingPopUp(true);
-    this.iconButton.owner.remove();
-    delete this.iconButton;
+    if (this.editorModeName === 'richText') {
+      this.editorPlugin.removeFormattingPopUp(true);
+      this.editorPlugin.removeIconButton();
+    }
```
merryman, line 3965: Can't this be handled via specialized commands that are either supported by the editor plugin or not? Or is there a scenario where the editorPlugin does support the formatting and icon button yet our editor mode is not set to rich text?

lively.halos/morph.js
```diff
@@ -2008,7 +2008,7 @@ export default class Halo extends Morph {
     const t = this.target;
     const topBar = this.topBar;
 
-    if (!t.isText) return;
+    if (!t.isText && !t.editorModeName === 'richText') return;
```
merryman, line 2011: Could the logic of this method be then moved into the editor plugin instead?

lively.ide/text/rich-text-editor-plugin.js
linusha: I will change that, but this got me thinking - why is the application of master components intertwined with the `onMouseDown` invocation of morphs? I.e., should we maybe think about dealing with this on an event level, so probably inside of the dispatcher, where then `onMouseDown` is invoked as well?
Not urgent in any case, and maybe I am also not seeing something important here, but relying on the correct invocation of onMouseDown seems like a footgun? I see that this kind of monkeypatching is not super great in any case, but overwriting onMouseDown and forgetting the super call will thus also brick the mater styling? :thinking:

lively.morphic/text/morph.js
linusha: No, you are right.

lively.halos/morph.js
linusha: Do you mean this whole method? Because I do not see how we can come out with some kind of check here in the halo, but we could of course check for the plugin and then invoke a method there.

merryman: Yes, the entire method basically, since all the code that follows after the check is only run in the presence of the rich text editor plugin, correct?

merryman (approved):

merryman (approved):

PR #664, opened 2023-01-27
🎨: only add text specific commands for interactive texts

### LINKED ISSUE #627: Commands requiring document crash when invoked on non selectable/document-free text morphs
As part of our unification of Label and Text there are now a bunch of Text morphs in the system that are unable to support the operations required by several commands that ship with text morphs by default.
One of these is for instance the ability to navigate the cursor via the arrow keys. But there are certainly others I have not yet checked. Solving this basically requires either: (1) Weeding out all the critical commands and making them aware not every text morph is configured to support them or (2) developing a mechanism that filters the set of commands for a text based on the document/text layout being present or not.

linusha:
Closes #627.

When going through the list of commands that are available on `Text`, it turned out, that they all only really make sense when one has an editable morph at ones hand. We briefly thought about extracting the commands into an EditorPlugin or the base EditorPlugin, but since the whole editing of `Text` is implemented via commands, and we have things like the world naming input line which require editing but do not have any plugin, and we did not want to have an IDE dependency to a new plugin in the text morph, we opted for this solution.

### DIFF
--- a/lively.morphic/text/morph.js
+++ b/lively.morphic/text/morph.js
@@ -1579,6 +1579,7 @@ export class Text extends Morph {
   }
 
   get commands () {
+    if (!(this.document && this.textLayout)) return [];
     return this.pluginCollect('getCommands', (this._commands || []).concat(commands));
   }
 


lively.morphic/text/morph.js
```diff
@@ -1579,6 +1579,7 @@ export class Text extends Morph {
   }
 
   get commands () {
+    if (this.readOnly) return [];
```
merryman, line 1582: We do have text morphs that are `readOnly`, yet still have a document + text layout present and make use of it. For instance in the *upon coming* component reconciliation, I sometimes set the browser source code editor to `readOnly = true` but still select/highlight ranges of code.
So I believe directly checking for the document is the better approach here.

linusha: Do you have an example for a command which you would like to execute in this state? I.e., where the browser source editor is set to be readonly?
In the case that one cannot edit text, de facto none of the commands make sense.

merryman (approved):

Issue #609, opened 2023-01-07
We need a Dock for iconified windows...


rickmcgeer:
In every Lively session I've ever had (for 10 years!), I wind up with iconified windows that are scattered all over the studio, and (further) it's hard to tell which iconified window is for what -- if I have three inspectors open, which is for what object?
I suggest that iconified windows be stored in rows at the bottom of the screen, similar to the taskbar in windows and the dock in MacOS, and, moreover, that the icon title either contain the name of the object or the object appears in a hover

merryman:
I am not entirely sure how the dock fixes that. How for instance does the dock fix the problem of what inspector is linked to which object in the scene? Is this not more deeply related to the way the inspector is implemented right now?

linusha:
@merryman I think @rickmcgeer might clarify this for us, but I am pretty sure that those are different feature ideas.

1. A dock at the bottom of the screen holding all iconified windows.
2. The titles of windows and thus of the iconified windows need to contain better information, i.e., the name of the inspected object for inspectors. This is definitely an issue and I noted that down with an emphasis on browser in the past (#378).

I am fully on board with 2. I would also propose that we maybe add an additional button to the inspector which `shows` the current target object? De facto the reverse of the "bind to new target button" we currently have?

---

As for the dock, I will think about this some more, but I think it is important that this does not take up screen estate when it is not currently used. Otherwise, I fear we are running in a situation where we take up too much screen space with all our tooling (topbar, sidebars, minimap,...)

linusha:
Until further notice, #670 solved this.

PR #670, opened 2023-01-31
Feature/new window switcher

linusha:
Here is a quick demo, what this feature can do:

https://user-images.githubusercontent.com/14252419/215834566-9e6b86e4-ae22-4fd6-afd1-ae03c08feeba.mp4

Namely:

- Show an overview of all open windows.
- Allow filtering of all open windows.
- Clicking on a window focuses and deminimizes this window.
- Windows can be closed from this overview.

The window switcher can be toggled by executing the `window switcher` command or via the window menu.
Additionally, the title of inspector and browser windows gets improved with the changes on this branch. 

@rickmcgeer imo, this solves #609. What do you think?


lively.ide/window-switcher.cp.js
```diff
@@ -92,6 +92,7 @@ const WindowPreview = component({
     extent: pt(100, 100),
     fill: Color.transparent
   },
+  { name: 'overlay', fill: Color.transparent, extent: pt(130, 130), nativeCursor: 'pointer' },
```
linusha, line 95: Why did you opt to insert a new morph for this? Wouldn't it be enough to just set the `nativeCursor` on `'window preview'` above? :thinking:

merryman (approved):

lively.ide/window-switcher.cp.js
merryman: That would work, if the previews themselves were plain textures. But they are html, and therefore happily carry over their respective css cursor styles, so the easiest way to get rid of this is to just slap a layer on top.

linusha: Aaah, yes! That makes sense!

Issue #689, opened 2023-02-06
Error when click component button in new project


kholbekj:
Error in interactive command browse and load component: TypeError: Cannot read properties of undefined (reading 'isCo...
TypeError: Cannot read properties of undefined (reading 'isComponentDescriptor')
    at Window.part (http://localhost:9011/lively.morphic/components/core.js:15:29)
    at Object.exec (http://localhost:9011/lively.ide/world-commands.js:1006:111)

merryman:
We are still working on implementing that feature. It is part of #476.

kholbekj:
Maybe a fun little thing for me to try to figure out how to do could be a nice "not implemented yet" warning? :) Always good to be able to distinguish expected failure from unexpected!

merryman:
You are welcome! We do have some parts in the code where we use a `nyi()` call. 
Take look at: `$world.execCommand("open browser", {moduleName: "src/resource.js", packageName: "lively.resources" });`
But we have not real consistent handling of this sort of thing in the system.

linusha:
Since the `nyi` PR is not really related to the fact that the component browser is missing at this time, I'll close here for now.

PR #692, opened 2023-02-08
🧰: add not yet implemented error to component browser

kholbekj:
https://github.com/LivelyKernel/lively.next/issues/689

### DIFF
--- a/lively.ide/studio/top-bar.cp.js
+++ b/lively.ide/studio/top-bar.cp.js
@@ -1,7 +1,7 @@
 import { Color, Rectangle, LinearGradient, rect, pt } from 'lively.graphics';
 import { config, touchInputDevice, TilingLayout, morph, Text, Polygon, Path, HTMLMorph, Ellipse, Morph, Image, Label, ShadowObject, Icon, component, ViewModel, part } from 'lively.morphic';
 import { Canvas } from 'lively.components/canvas.js';
-import { Closure, string, obj, arr } from 'lively.lang';
+import { Closure, string, obj, arr, fun } from 'lively.lang';
 import { resource } from 'lively.resources';
 import { CommentBrowser } from 'lively.collab';
 import { once, connect, disconnect, signal } from 'lively.bindings';
@@ -11,14 +11,6 @@ import { UserUI } from 'lively.user/morphic/user-ui.js';
 import { SystemTooltip } from 'lively.morphic/tooltips.cp.js';
 import { RichTextPlugin } from '../text/rich-text-editor-plugin.js';
 import { WorldMiniMap } from '../world-mini-map.cp.js';
-
-function nyi (msg, strict = false) {
-  let error = new Error(`Not yet implemented ${msg}`);
-  $world.logError(error);
-  if (strict) {
-    throw error;
-  }
-}
 
 class SelectionElement extends Morph {
   static get properties () {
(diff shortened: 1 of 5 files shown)


lively.ide/studio/top-bar.cp.js
```diff
@@ -12,6 +12,14 @@ import { SystemTooltip } from 'lively.morphic/tooltips.cp.js';
 import { RichTextPlugin } from '../text/rich-text-editor-plugin.js';
 import { WorldMiniMap } from '../world-mini-map.cp.js';
 
+function nyi (msg, strict = false) {
+  let error = new Error(`Not yet implemented ${msg}`);
```
linusha, line 16: I think personally I'd prefer switching the message around so that, e.g., in this case it would read: "Component Browser not yet implemented". No hard feeling here though, just a thought.

linusha (commented):
Nicely done, thanks!

However, I think we can still do one better: One problem with `nyi` is, that there are several implementations all over the place because we thought we had no canonical implementation.

We have a package called `lively.lang` which is comparable in form and function to utility libraries like `lodash`. In there, there exists an `notYetImplementd()` function, but somehow it never got used and we got multiple implementation all over the place.

I propose the following, if you are up to that:

1. We move your implementation into `lively.lang` (`$world.execCommand("open browser", {moduleName: "function.js", packageName: "lively.lang", codeEntity: "notYetImplemented"});`), since yours is nicer and more flexible. You can keep `nyi` or the current name in `lively.lang`, as you prefer. In case you rename just check the export structure of the package as well.
2. We import that function in the topbar.
3. If you like, however that one is optional for merging this, you could eliminate other implementation of `nyi` (from a quick glance there is one in `layout.js` and I believe in `resource.js` and instead import your new lively.lang one.

lively.ide/studio/top-bar.cp.js
kholbekj: You are very right, it was slightly copy-pasta from elsewhere, I'll fox 🦊

kholbekj:
It's great feedback, thanks! I'll jump on that when I got a bit of time!

kholbekj:
There's one more occurence that raised a few questions : `lively.2lively/interface.js` has one too. What is 2lively? Should I update that one too?

linusha:
Ah, yes, I did not see that one initially. Exchanging that as well would be great! 

lively2lively is, or was, a messaging system between mulitple lively instances that are connected to the same server. In the past (the lively project had multiple people as a whole research team working on it at some point), there was more emphasis placed on the server, with clients connecting to that server and then all people could e.g., share parts (prepared objects with attached behavior, kind of what modern web tech calls components, chat with each other (and execute JS code on all other connected systems :upside_down_face:).
For resource reasons, we currently strayed quite far from this approach, I am currently working on GitHub integration in lively so that we can just use that for authentication and sharing of stuff and the server is more or less just an interface from your browser to your local file system.

So the golden days of lively2lively are gone, but it would be cool to polish that up. We loosely thought about extracting the L2L server, so that one can spin that up somewhere and then lively sessions could connect to that without the rest of the server being impacted. Then, we could get chatting, code sharing, etc. going again.

If that would be something that interests you, just hit us up! 

P.S. I will write you some more about the project and what I know from its past in matrix. I hope to get to it over the weekend.

kholbekj:
Thanks for the context 😍

Well fix that one as well, then!

kholbekj:
Lots of autoformatting on that last one, but I assume that's ok?

linusha:
> Lots of autoformatting on that last one, but I assume that's ok?

Nice! Yeah, that's alright. When editing really old files (i.e., files that have not changed for a long time), the auto-lint will kick in, since they were never linted. When there are more involved changes, a commit with the auto-formatting changes alone is also fine, to make it easier to spot the relevant changes in a commit.

But this is great the way it is! :slightly_smiling_face:  Thanks! :tada:

linusha (approved):

PR #702, opened 2023-02-10
🎨: allow to hide scrollbars in `Text`

### LINKED ISSUE #700: `hideScrollbar` property does not work correctly on `Text`
This is due to the separate scrollLayer that `Text` with a `Document` has.
We just need to adapt the renderer in a way that this property gets propagates to the subnodes as well.
We need to make sure that the reversal works too.

linusha:
Closes #700.

### DIFF
--- a/lively.morphic/morph.js
+++ b/lively.morphic/morph.js
@@ -544,7 +544,13 @@ export class Morph {
         group: 'styling',
         type: 'Boolean',
         isStyleProp: true,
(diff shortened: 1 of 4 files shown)


merryman:
Just to clarify: This property *only* works on text morphs right? When I apply this to normal morphs with clipMode set to `'scroll'` or `'auto'` it has not effect.

merryman (approved):

PR #717, opened 2023-02-17
Fix/prevent infinite render loops

### LINKED ISSUE #712: New Layout(?) procedures are causing multiple infinite loops in galyleo-dashboard
A good example (among others) is in https://github.com/engageLively/galyleo-dashboard/issues/158.
I suspect that this isn't a bug, but a change in layouting code that early developers and applications (like Galyleo) will have to adapt to.  No problem, but we do need guidance and any guidance is appreciated.

merryman:
Closes #712.

### DIFF
--- a/lively.morphic/rendering/renderer.js
+++ b/lively.morphic/rendering/renderer.js
@@ -890,8 +890,7 @@ export default class Renderer {
    * @param {Text} morph
    */
   removeTextSpecialsFromDOMFor (node, morph) {
-    const cursorNode = node.querySelector('.newtext-cursor');
-    cursorNode.remove();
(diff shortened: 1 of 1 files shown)


lively.morphic/rendering/renderer.js
```diff
@@ -889,7 +889,7 @@ export default class Renderer {
    */
   removeTextSpecialsFromDOMFor (node, morph) {
     const cursorNode = node.querySelector('.newtext-cursor');
-    cursorNode.remove();
+    if (cursorNode) cursorNode.remove();
```
linusha, line 892: node.querySelector('.newtext-cursor')?.remove()

linusha (approved):

Issue #581, opened 2022-12-10
Fonts aren't loading properly in frozen world


rickmcgeer:
This may be an overlap with #571 , in which case I apologize...

Here's the dashboard in the lively-next development system:
![Untitled drawing (1)](https://user-images.githubusercontent.com/656259/206871383-1dc35a3f-2afb-42d2-95fd-484bdf7eff37.png)

And here's the frozen world:
![Untitled drawing](https://user-images.githubusercontent.com/656259/206871393-6b0c61e0-4b27-4728-a554-23da6214eb36.png)

For Galyleo, I'm just going to use a generic font (Sans-Serif) as the font attribute for each of the affected components, which should let us do a beta deploy.  However, this is something I'm sure you guys want to fix.

PR #719, opened 2023-02-17
🎨: wait for $world on ensureFont()

merryman:
Fixes an issue where `ensureFont()` would just terminate without loading the font if invoked before `$world` was initialized. (happens quite a bit in bundles).


linusha:
@merryman does this solve #581?

merryman:
Pretty sure it does!

linusha (approved):
Pretty sure we want to revisit this when dealing with the font management as a whole. But if it solves an acute problem for now, let's do it! :rocket:

PR #726, opened 2023-02-28
🧰: move window into view that is selected in window switcher

merryman:
This fixes an issue when moving between large and small displays (like unplugging your laptop from a large external display).
In those cases there may be numerous windows that are now outside of the window bounds. The window switcher is a great way to get a hold of those windows, but up till now would not move them into view once selected.
This fixes that.


linusha:
I might just do not get what is happening here, but this does not change the code at all? :thinking: Isn't this just removing `this` by inserting one line of destructuring assignment?

merryman:
True, I must have smuggled it in as part of #676. my bad.

linusha:
Yes, this was already in bd0a25fc5e65e2cc4f7572be86197b45a46b42a8.

PR #727, opened 2023-03-01
❄️: ensure to trigger loading indicator after loading screen is mounted

### LINKED ISSUE #725: LoadingIndicator is not always visible in loading screen
As the title states, the loading indicator is often not visible. This is not always the case. I believe it is always missing when one loads lively in a given tab for the first time. Subsequent loads in the **same tab** will have the loading indicator.
We should fix this, as it is sometimes hard to grasp if one is stuck in the loading screen or if it just takes a while.

merryman:
Closes #725.

### DIFF
--- a/lively.freezer/src/loading-screen.cp.js
+++ b/lively.freezer/src/loading-screen.cp.js
@@ -31,7 +31,7 @@ export class WorldLoadingScreen extends Morph {
     return `<script> window.AUTH_SERVER_URL; </script>`;
   }
 
-  async onLoad () {
+  async activate () {
     if (lively.FreezerRuntime) {
       const progressBar = this.get('package loading indicator');
       const cssLoadingScreen = this.get('css loading screen');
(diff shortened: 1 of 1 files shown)


linusha:
This is more of a general question, but when does the `onLoad` "hook" gets executed - I looked around the source and it looks to me like it happens when the module gets loaded, is that correct?
In any case - why did it not suffice to make the loading indicator work here? :thinking:

merryman:
The semantics of `onLoad()` are a little bit weird due to the complex relationship between the two main morph creations schemes in lively. Those are either invocation of the constructor like: `new MorphClass()`, `morph(...)`, `part(...)` or via the deserialisation from a snapshot. 
Note that **we do not invoke the constructor when a morph gets deserialised!!** I believe this is something that we do not mention often enough in our documentation about serialisation.
So the purpose of `onLoad()` here is to provide a lifecycle callback that always gets called after either deserialisation or constructor initialization has taken place (think of it as a universal constructor).
In it, you can set up ephemeral objects and other external state that is needed to get the morph going.

We do allow `onLoad()` to also be asynchronous, which was the case in this example. That created a race condition between the font loading and the `onLoad()` itself, which would initialise the loading indicator and the loading screen in different orders.

linusha (approved):

Issue #668, opened 2023-01-31
Focus is very brittle


linusha:
Often, the focus does not lie with the element one would suspect. This leads to e.g., keystrokes not registering correctly, which is frustrating.

To better investigate this, we want to add a small widget that always displays the currently focused element.

PR #732, opened 2023-03-03
🎛️: stop muller columns from stealing active text focus

merryman:
Fixes a common issue where muller columns inside the system browser would steal the focus when hovered, while a text morph was still holding onto the focus in order to receive text input. I believe this is one of the main causes for unfocused filtered list prompts.

Should fix some of #668


linusha (changes requested):
This is unfortunately incompatible with the filter capabilities of the muller columns.

linusha (approved):

PR #753, opened 2023-03-08
🎨: fix style property cache reset mechanism

merryman:
Glancing over #739, I discovered that the reset mechanism for the style properties cache itself never really worked. This fixes that.


lively.morphic/helpers.js
```diff
@@ -71,7 +71,7 @@ export function addClassMappings (mapping) {
 // possible solution: store the modules for each of the classes we cache the style props for and clear the cache when a change in the module is sent around from lively.notifications
 
 export function clearStylePropertiesForClassesIn (moduleId) {
-  CachedStyleProperties.forEach(([klass, props]) => {
+  CachedStyleProperties.forEach((props, klass) => {
```
linusha, line 74: That `Map >> forEach` is this way around is peak JS :shrug:

linusha (approved):

PR #761, opened 2023-03-13
Fix/browser status panel covered controls

merryman:
Fixes an issue where the control buttons inside the browser's status panel would be covered by some elements that are currently invisible.

Brefore:
unresponsive-buttons

After:
responsive-buttos

### DIFF
--- a/lively.ide/js/browser/ui.cp.js
+++ b/lively.ide/js/browser/ui.cp.js
@@ -459,7 +459,203 @@ export class PathIndicator extends Morph {
   }
 }
 
-// b = part(SystemBrowser)
+const MetaInfoContainerExpanded = component({
+  type: PathIndicator,
+  master: FileStatusDefault,
+  isLayoutable: true,
+  clipMode: 'hidden',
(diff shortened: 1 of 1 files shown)


lively.ide/js/browser/ui.cp.js
```diff
@@ -459,7 +459,203 @@ export class PathIndicator extends Morph {
...
+  }]
+});
+
+const MetaInfoContainer = component(MetaInfoContainerCollapsed, {
+  // master: {
+  //   states: {
+  //     expanded: MetaInfoContainerExpanded,
+  //     collapsed: MetaInfoContainerCollapsed
+  //   }
+  // }
```
linusha, line 662: Can we remove this? Also, it looks like the whole derivation can go, no?

linusha (changes requested):
I am quite certain that this is the case, but just to make sure: The label holding the current package name (the one underlined) does also not get blocked anymore, right?

linusha (approved):

PR #782, opened 2023-03-21
Enhancement/double click in browser

linusha:


lively.components/tree.js
```diff
@@ -480,6 +480,10 @@ export class Tree extends Text {
     if (onMouseDown) onMouseDown(evt);
   }
 
+  onDoubleMouseDown (evt) {
+    signal(this, 'doubleMouseDown', this.selectedNode);
+  }
+
```
merryman, line 486: Seems we do not really need this override.

PR #820, opened 2023-04-25
🧰🖌️: allow drag-creation of morphs not only from top-left

linusha:
Previously, we could only create shapes with the top-bar tooling from by dragging from the top-left to the bottom-right of the shape to be created. However, we could create grey selections in the other direction as well.

This PR reuses the strategy from selections for the other shapes in order to allow easier creation of shapes.

### DIFF
--- a/lively.graphics/geometry-2d.js
+++ b/lively.graphics/geometry-2d.js
@@ -582,6 +582,10 @@ export class Rectangle {
       this.x + (r.x * this.width), this.y + (r.y * this.height), r.width * this.width, r.height * this.height);
   }
 
+  scaleBy (scaleX, scaleYOrUndefined) {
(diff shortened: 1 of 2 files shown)


lively.graphics/geometry-2d.js
```diff
@@ -582,6 +582,10 @@ export class Rectangle {
       this.x + (r.x * this.width), this.y + (r.y * this.height), r.width * this.width, r.height * this.height);
   }
 
+  scaleByFactor (num) {
+    return new Rectangle(this.x, this.y, this.width * num, this.height * num);
+  }
+
```
merryman, line 588: I would prefer something that is more similar and capable to the function found in `Point`:
```suggestion
  scaleBy (scaleX, scaleYOrUndefined) {
    return new Rectangle(this.x, this.y, this.width * scaleX, this.height * (scaleYOrUndefined || scaleX));
  }

```

merryman (changes requested):

PR #821, opened 2023-04-25
🪨: silence errors when http-resource results in 404

linusha:
In 06fa501d4f85ebd3cea928536eaa7f3c507ec8ad we added a check for the .gitignore file in a package, which works by using `exists` calls. As for the core packages, usually no .gitignore exists, this lead to a lot of 404 errors in the console.
This should resolve this issue and I could not see any downsides - but please verify and correct me, might very well be that I overlooked something.

Screenshot from 2023-04-25 16-26-58


merryman:
I would refrain from making the silencing of errors right within the `makeRequest ` method. I have found it quite helpful actually to check the console every once in a while in case a particular file was really not present that I expected to be able to read. How about placing the `try { } catch` in the `exists()` method?

PR #824, opened 2023-04-26
🎨: add cursor position to debug view of `Text`

### LINKED ISSUE #786: Debug Mode should show current column of cursor
**Describe the Feature**
Besides the currently visible rows, the debug layer should also show the current column of the cursor.

**Graphics**
image

linusha:
image

Closes #786.


merryman:
Nice, but we should do something about the fact that the debug info now covers a significant portion of the top 2 lines. Can we set the background of the `.debug-info` to something a little transparent? Like `rgb(255 255 255 / 80%)`.

merryman (approved):

PR #838, opened 2023-05-04
🎨: implement match based bindings

merryman:
Allows us to *finally* define dynamic bindings that (potentially) match multiple morphs and also are updated whenever the view changes.


linusha (commented):
~~@merryman could you provide me with a screenshot/video of this in action? I somehow cannot get it to work but that is probably just me missing something obvious.~~

Also, when I open a component in not-lively mode and drop a submorph, an error gets logged in the console. Shouldn't happen I guess?

image

linusha (approved):

PR #859, opened 2023-05-24
🧰: ensure new empty js modules are in esm format

### LINKED ISSUE #852: Warning for discarded changes upon first save in a created module
**Describe the bug**
When one creates a new file in the system browser, upon the first save one gets warned that changes will be discarded. This does not actually happen, but we should not show the prompt

**To Reproduce**
Just create a new module, insert some text and save.

**Setup**

- Linux

**Version**: d6d294003914fdf1a2720de9b49b0ce61a6c9d23

merryman:
Closes #852.

### DIFF
--- a/lively.ide/js/browser/index.js
+++ b/lively.ide/js/browser/index.js
@@ -2384,7 +2384,8 @@ export class BrowserModel extends ViewModel {
     }
     if (!name) return;
     let dirPath = joinPath(dir, name);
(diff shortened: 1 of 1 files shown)


linusha (approved):
I can confirm that this works, but I do not understand why that is the case.
Could you elaborate why this is the appropriate fix before we merge this?

PR #863, opened 2023-05-31
🛠️: skip minification when building artifacts during install

linusha:


merryman (commented):
So when *do* we actually minify now? just after manually invoking build-artefacts?

linusha:
Invocation of the various make targets also skips that. We could think about changing that when we are at a point where we again deploy instances semi-often at least.
Running the `npm run build-landing-page` and `npm run build-loading-screen` commands will minify the code of these two. In all other instances, we skip that and at this point I would argue that maybe it makes sense to reverse all this and just make the minification for the landing page and the loading screen opt-in, since, as mentioned above, that is irrelevant in nearly all cases and just takes time.

merryman (approved):

PR #876, opened 2023-06-02
📂: add option to show `git diff-files --stat` when saving project

linusha:


merryman:
Creating a new project seems to be broken with these changes. @linusha Have you tried just creating a new empty project? In my case it fails with "Error saving the project".

lively.project/prompts.cp.js
```diff
@@ -453,7 +462,17 @@ export const SaveProjectDialog = component(SaveWorldDialog, {
...
+        name: 'diff button',
+        extent: pt(449.5, 27),
+        submorphs: [{
+          name: 'label',
+          textAndAttributes: ['Show Summary of Changes (Advanced Operation)', null, '', {
+            fontFamily: 'tabler-icons',
+            fontWeight: '900'
+          }, ' ', {}]
+        }]
+      })]
```
merryman, line 475: Can we replace with with a dark system button? It will look better when placed on the dark prompt.

merryman (changes requested):

merryman:
> > Creating a new project seems to be broken with these changes. @linusha Have you tried just creating a new empty project? In my case it fails with "Error saving the project".
> 
> Yes, works flawlessly for me, with and without github syncing. https://github.com/linusha/canIsave Any more pointers as to what you did? Besides, I heavily doubt that this has something to do with the changes in here, or can you reproduce that something that works on main does not work here?

Never mind. Now works for me. It must have been something about my setup. Maybe old stuff inside the local storage?

lively.project/prompts.cp.js
linusha: image

merryman (approved):

PR #877, opened 2023-06-02
📂🧰: only bind and check projects against (shared) commits on main

linusha:
This intends to make it easier to work with feature branches in combination with projects while developing the core of lively.next. Previously, we would bind projects against the currently checked out commit. When developing on a feature branch, especially when using `git rebase`, this often lead to situations where projects would not open anymore due to incompatible versions.

We now only bind against commits on `main`.
The check for a compatible version also checks the relationship between the latest commit that is shared with `main` and the currently checked out branch and the version specified in the `package.json` of the project.

The following scenarios are possible:
1. Both are the same and we are good to go.
2. The latest shared commit is newer than the previous commit from `main` we bound the project against. In this case, we update the bound version in the `package.json` of the project.
3. If the version of `main` in the `package.json` is newer than the latest shared commit of the running lively and `main`, we need to update our lively. Previously, we aborted the loading of the project in these cases. Now, we only display an explicit warning, but continue loading the project. This is an additional change that intends to make quickly testing stuff easier.

@merryman I tested this and I think it works as expected, but a thorough review for content and not "only" code would be appreciated!

### DIFF
--- a/lively.project/project.js
+++ b/lively.project/project.js
@@ -69,7 +69,7 @@ export class Project {
       version: '0.1.0'
     };
 
-    if (bindVersion) VersionChecker.currentLivelyVersion().then(version => this.config.lively.boundLivelyVersion = version);
+    if (bindVersion) VersionChecker.currentLivelyVersion(true).then(version => this.config.lively.boundLivelyVersion = version);
   }
 
   static async listAvailableProjects () {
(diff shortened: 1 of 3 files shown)


lively.project/project.js
```diff
@@ -152,27 +152,22 @@ export class Project {
...
     const checkLivelyCompatability = await loadedProject.bindAgainstCurrentLivelyVersion(loadedProject.config.lively.boundLivelyVersion);
 
-    let loadingCanceled = false;
     switch (checkLivelyCompatability) {
       case 'CANCELED':
       case 'OUTDATED': {
-        loadingCanceled = true;
-        await $world.inform('Due to incompatible lively versions, you cannot open this project. This session will now close. This probably means that your version is older than the one the project requires.');
-        window.location.href = (await Project.systemInterface.getConfig().baseURL);
+        await $world.inform('The required lively version of this project conflicts with the running one.', { additionalText: 'You can proceed with OK, but be aware that some expected behaviour might differ or not work.' });
```
merryman, line 158: formatting-issue
Please fix the formatting of the additional text. This looks to much out of place for my taste.

merryman (changes requested):

merryman (approved):

PR #878, opened 2023-06-02
Feature/asset bundling

linusha:


lively.project/helpers.js
```diff
@@ -0,0 +1,11 @@
+export function projectAsset (fileName) {
+  // Uses Error() to generate the stacktrace from this function being called.
+  // Afterwards, we find out from within which project we were called.
+  // This is the project of which we need to construct the asset path!
+  const callStack = (new Error()).stack;
+  const callingProject = callStack.match(/\(.*\/local_projects\/(?<callingProject>[a-zA-Z_-]+)\/.*\.js/m)?.groups.callingProject;
+  if (lively.FreezerRuntime) {
+    return './assets/' + fileName;
+  }
```
merryman, line 9: Can we move this to the top of the method? For what its worth, we save the exception object and the string matching, when its not nessecary.

merryman (changes requested):

merryman (approved):

PR #893, opened 2023-06-12
Improvements for Projects, esp. `flatn` resolution support

linusha:


linusha:
@merryman when we merge this, we need to adapt a) all local folder structures so that they use double-dash instead of single dash to separate owner and name of the project and b) inside of each `package.json` for a project, the name needs to be `<owner--name>` as well! I already did that for the partsbin repository.

merryman:
I emptied the `local_projects` folder, to respect the new folder structure and naming convention.
Now it appears I can not create a new project from remote any more:
cannot-init-project

merryman (changes requested):

merryman (approved):

PR #900, opened 2023-06-21
Fix/side bar performance issues

merryman:
Fixes some performance issues with the renderer and widgets that were mainly affecting the properties side bar.

### DIFF
--- a/lively.morphic/morph.js
+++ b/lively.morphic/morph.js
@@ -953,13 +953,13 @@ export class Morph {
     for (const key in properties) {
       try {
         const descr = properties[key];
-        if (
-          descr.readOnly ||
-        descr.derived ||
-        descr.isComponent ||
-        (descr.hasOwnProperty('serialize') && !descr.serialize) ||
-        obj.equals(this[key], defaults[key])
-        ) continue;
+        if (descr.readOnly ||
+            descr.derived ||
+            descr.isComponent ||
+            (descr.hasOwnProperty('serialize') && !descr.serialize) ||
+            obj.equals(this[key], defaults[key])) {
+          continue;
+        }
         propsToSerialize.push(key);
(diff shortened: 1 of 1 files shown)


lively.morphic/layout.js
```diff
@@ -545,7 +545,12 @@ export class TilingLayout extends Layout {
    */
   get hugContentsVertically () {
     if (this.wrapSubmorphs && this.axis === 'column') return false;
-    if (!this.resizePolicies.some(([_, { height }]) => height === 'fixed')) return false;
+    for (let m of this.layoutableSubmorphs) {
+      const h = this._resizePolicies.get(m)?.height;
+      if (!h) continue;
+      if (h === 'fill') return false;
+    }
+    // if (!this.resizePolicies.some(([_, { height }]) => height === 'fixed')) return false;
```
linusha, line 553: I think this can go?

lively.components/list.js
```diff
@@ -1870,29 +1873,36 @@ export class DropDownListModel extends ButtonModel {
...
+      } else {
+        const { items } = listMorph;
+        let item = listMorph.find(sel);
+        if (!item && typeof sel === 'string') {
+          item = items.find(ea => ea.string === sel);
+        }
+        if (!item) return super.onRefresh(prop);
 
-      if (prop === 'selection') this.adjustLableFor(item);
+        this.adjustLableFor(item);
```
linusha, line 1900: This now done more than previously, was this intended? The condition above is less strong than the old one, no?

merryman: Slightly so, but really only if the condition `this.updateListMorphIfNeeded() || prop === 'selection' || prop === 'items'` is true, which is not super more often than before.

linusha (approved):

PR #908, opened 2023-06-29
Ship lively with extensive default font bundle and add support/tooling for project specific fonts

linusha:
After pulling this update, running `make artifacts` is required! (This is always recommended, but this time even more so.)

With this PR, we will ship around 30 fonts that can be used inside of `lively.next` that are statically available and can also be bundled with frozen applications.

We also introduce the possibility to have `index.css` and `fonts.css` files in `lively.projects`. Those will also be bundled and it is ensured that they are applied when the project is loaded.

Inside of `index.css` one can put any CSS. `fonts.css` is not meant for manual editing!

Instead, inside of the text controls in the properties panel, a new tool, the "font manager" can be opened that allows to upload custom fonts in the `woff2` format. Those fonts will also be automatically loaded at runtime inside of `lively.next` and will be bundled as well.

image

**However, this PR deprecates the usage of `fontFamily` values which are not fonts that come with `lively.next` or are part of the custom fonts of a project. This especially means that values such as `Sans-Serif` will no longer be supported!**

@rickmcgeer I will create PRs for you for fez-intro and the GalyleoDashboard. However, please take note of these changes for future endeavors! :slightly_smiling_face:


lively.morphic/text/morph.js
```diff
@@ -611,13 +612,14 @@ export class Text extends Morph {
...
-        values: DEFAULT_FONTS.map(f => f.name),
+        values: availableFonts().map(f => f.name),
         defaultValue: 'IBM Plex Sans',
         isStyleProp: true,
         isDefaultTextStyleProp: true,
         after: ['defaultTextStyle'],
         set (fontFamily) {
-          this.setProperty('fontFamily', sanitizeFont(fontFamily));
+          document.fonts.load(`${this.fontStyle} ${this.fontWeight} 12px ${fontFamily}`)
+            .then(() => this.setProperty('fontFamily', sanitizeFont(fontFamily)));
```
merryman, line 622: I am a little worried about placing the asynchronous part inside a setter. This can lead to unexpected behaviour since we usually assume that setting a property is immediately reflected in the state of the object.
Can we maybe defer the rendering instead of deferring the property setting?

lively.morphic/text/morph.js
```diff
@@ -1015,47 +1026,39 @@ export class Text extends Morph {
...
-        console.log('checking', this.name);
-        setTimeout(check, 500);
-        return;
-      }
-      p.resolve(false);
-    };
-    check();
-    return p.promise;
+    const usedFonts = this.usedFonts();
+    const success = await Promise.all(usedFonts.map(fontString => this.fontMetric.supportedFontCache.has(fontString) || document.fonts.load(fontString)));
```
merryman, line 1059: This code assumes that a font that is both missing in the `supportedFontCache` and is false with regards to `document.fonts.check()` is not listed in any css/font face in the dom. Is this guaranteed to be the case? I was wondering if maybe the same condition can be the result of incomplete font loading, etc.. This is an issue that will never arise in practice when working within lively.next, but I could imagine this being an issue in frozen apps?

lively.ide/world.js
```diff
@@ -1535,3 +1541,36 @@ export class LivelyWorld extends World {
...
+      else this.fileActions[file] = [this.fileActions[file], cB];
+    } else { this.fileActions[file] = cB; }
+  }
+
+  unregisterFileActions (file) {
+    this.fileActions[file] = null;
+  }
+
+  async onFileSave (savedFileData) {
+    if (!this.fileActions[savedFileData.resource]) return;
```
merryman, line 1568: Can we also make this match regular expressions instead of the literal file names? That would allow us to perform general action on when a css file changes, such as updating it in all browsers.

lively.project/project.js
```diff
@@ -198,12 +200,40 @@ export class Project {
...
+  /**
+   * @param {string} projectUrl
+   * @param {boolean} forProject - Whether to install CSS for an opened project or a dependency
+   * @param {object} opts
+   * @param {object} projectRef - Only important when forProject is true. Reference to the Project that the CSS belongs to, so that we can access its instance later.
+   */
+  static async installCSSForProject (projectUrl, forProject, opts, projectRef) {
+    const indexCSSResource = resource(projectUrl).join('index.css');
+    let indexCSS = await indexCSSResource.read();
+    if (forProject)indexCSS = `@import '/local_projects/${opts.repoOwner}--${opts.name}/fonts.css';\n` + indexCSS;
```
merryman, line 221:
```suggestion
    if (forProject) indexCSS = `@import '/local_projects/${opts.repoOwner}--${opts.name}/fonts.css';\n` + indexCSS;
```

lively.project/project.js
```diff
@@ -198,12 +200,40 @@ export class Project {
...
+    else indexCSS = `@import '/local_projects/${opts.name}/fonts.css';\n` + indexCSS;
+    if (forProject) addOrChangeCSSDeclaration(`CSS-for-project-${opts.name}`, indexCSS);
+    else addOrChangeCSSDeclaration(`CSS-for-dependency-${opts.name}`, indexCSS);
+    if (forProject) {
+      const updateProjectCSS = async () => {
```
merryman, line 226: Can we merge this into a single `if {} else {}`?

lively.ide/studio/font-manager.cp.js
```diff
@@ -0,0 +1,762 @@
...
+    this.ui.dragAndDropArea.borderWidth = 7;
+  }
+
+  onNativeDragleave () {
+    this.ui.dragAndDropArea.borderWidth = 2;
+  }
+
+  onNativeDragend () {
+    this.ui.dragAndDropArea.borderWidth = 2;
+  }
```
merryman, line 174: Better to be handled via a master component tbh.

merryman (changes requested):

lively.ide/world.js
linusha: Yes, we can certainly do that. I would however prefer to do that separately and not as part of this PR, is that okay with you?

Issue #912, opened 2023-07-02
Timeouts in build


rickmcgeer:
**Describe the bug**
The build on my lively project gets through minifying the source files, but then hits a timeout with this error:

```
100%file:///home/rick/lively.next/lively.lang/promise.js:125
          ? reject(new Error('timeout'))
                   ^

Error: timeout
    at Timeout.<anonymous> (file:///home/rick/lively.next/lively.lang/promise.js:125:20)
    at listOnTimeout (node:internal/timers:569:17)
    at process.processTimers (node:internal/timers:512:7) {
  code: 'PLUGIN_ERROR',
  plugin: 'rollup-plugin-lively',
  hook: 'generateBundle'
}
```

**To Reproduce**
In local_projects

1. Clone the galyleo-dashboard repo: https://github.com/engageLively/galyleo-dashboard
2. npm run build

**Screenshots**

**Setup**
Please delete non-applicable items!

- Windows

**Version**: Please paste the lively.next commit on which the problem occurred here (use the copy button of the Version Checker in the bottom left corner).
5cecb8d2030617b92320cefc9ea4e63439e36630

rickmcgeer:
On a Mac, it's failing in the same spot (after minification) with a different error:

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0x103c917b5 node::Abort() [/usr/local/bin/node]
 2: 0x103c919a5 node::OOMErrorHandler(char const*, bool) [/usr/local/bin/node]
 3: 0x103e14dec v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
 4: 0x103fd9685 v8::internal::Heap::FatalProcessOutOfMemory(char const*) [/usr/local/bin/node]
 5: 0x103fd7f20 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/usr/local/bin/node]
 6: 0x103fc9d9a v8::internal::HeapAllocator::AllocateRawWithLightRetrySlowPath(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [/usr/local/bin/node]
[...]

PR #915, opened 2023-07-04
❄️: increase the timeout to wait for the minification to finish

merryman:
Hopefully deals with future cases of #912.


linusha:
How sure are you, that this is the cause? I mean I can see why it would fit with the error message we observed.
I ask because we could also run an experiment (build X times, count success, change timeout, do it again), but if you are reasonably sure we can also save some :polar_bear: for now and just proceed.

merryman:
I could not really reproduce the timeouts any more, so I will close this for the time being.

linusha:
Could not be reproduced for a long time.

PR #922, opened 2023-07-06
Enhancement/update lively

linusha:
This allows to update lively.next from within the GUI web environment when one is on the `main` branch and there are new changes available on `origin/main`.
The feature is intended for people that are mainly using lively as means to author their own tools, as opposed to people that are mainly working on the lively core. Thus, we can limit this functionality to main and also expect that people do not have outstanding stashes inside of their local lively core repository.

When an update is available, a hint is shown in the VersionChecker. Pressing that will trigger `update.sh` via the L2L Shell Interface. We pull new changes, rebuild the loading screen and landing page and in the end restart the server.

Restarting the server works by wrapping the old `start.sh` as `start-server.sh` inside of a new `start.sh`, that traps the TERM signal and restarts `start.sh` when lively gets stopped via this signal.

When the server is back up, we prompt the user to reload the page via a prompt.

---

This PR also deprecates the docker setup we previously had. While we still think a working docker setup would be a good idea, the one we currently had did not provide what one would have wanted, namely a really seamless operation without need for any dependencies locally, where just using docker on-board means suffices for running lively on all host platforms. The old setup did not provide this: As we wanted compatibility between docker-lively and native-lively, we mounted all dependencies into the container, and blowing up the interface one needed to interact with it, since we could not just set a uniform entry point for the `Dockerfile`.
The reason this came up now is that since that setup ran single scripts in containers via a Makefile, the above laid out approach to restart lively from within lively did not work, as first the container would stop and then the Makefile target would terminate.


lively.ide/studio/version-checker.cp.js
```diff
@@ -65,8 +69,22 @@ class VersionChecker extends Morph {
...
-  onMouseDown (evt) {
+  async updateLively () {
+    let li;
+    once(L2LClient.default(), 'onReconnect', async () => {
+      li.remove();
+      await $world.inform('Press OK to reload this page and finish the update.');
+      location.reload();
+    });
+    li = $world.showLoadingIndicatorFor($world, 'Updating lively');
+    runCommand('./../update.sh', { l2lClient: ShellClientResource.defaultL2lClient });
```
merryman, line 80: Cant we just bind this to `VersionChecker.cwd()` as well? Here we implicitly assume the default cwd will always be the `lively.server`, and if that will change in the future this code will break out of the blue.

lively.ide/studio/version-checker.cp.js
```diff
@@ -161,10 +179,20 @@ class VersionChecker extends Morph {
...
-    status.value = ['Version: ', {}, `[${version}]`, { fontWeight: 'bold' }, ' (Please update!)'];
-    this.updateShownIcon('behind');
+  async showBehind (version) {
+    const cwd = await VersionChecker.cwd();
+    const currentBranchCmd = 'git rev-parse --abbrev-ref HEAD';
+    let currBranch = await runCommand(currentBranchCmd, { cwd }).whenDone();
+    currBranch = currBranch.stdout.replace('\n', '');
+    const { status, updateButton } = this.ui;
+    if (currBranch === 'main') {
+      updateButton.visible = updateButton.layoutable = true;
```
merryman, line 189:
```suggestion
      updateButton.visible = updateButton.isLayoutable = true;
```

lively.ide/studio/version-checker.cp.js
```diff
@@ -161,10 +179,20 @@ class VersionChecker extends Morph {
...
+      status.value = ['Press Update icon to update!', { fontWeight: 'bold' }];
```
merryman, line 190: I dont know why, but it sounds a little.... meta?

update.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+echo "💾 Saving local changes if any."
+git stash
+
+echo "🪢 Retrieving the latest lively.next version."
+# Assumes that one runs this on `main` branch. Inside of lively, we ensure that the GUI trigger is not shown in other cases.
```
merryman, line 7: Isnt it better to be safe than sorry? Why not check if we really are on main here.

update.sh
```diff
@@ -0,0 +1,26 @@
...
+git pull
+
+echo "📦 Installing latest version of lively.next."
+./install.sh
+
+# If one has other stashes, they are popped regardless!
+echo "💾 Restoring local changes if any."
+git stash pop
```
merryman, line 15: So if I have some super outdated stash at 0 that conflicts with the current version of lively.next, we basically break the repo here? Can we add a check that only pops if the above stash really stashed files?

merryman (changes requested):

update.sh
linusha: Done

lively.ide/studio/version-checker.cp.js
linusha: Done

PR #940, opened 2023-07-17
🛠️🧰: enforce `LF` line-endings in repository and system browser

### LINKED ISSUE #187: Unify linebreaks between platforms on text enter/render/save
When I use the Object Editor (or workspace) and select part of a line, the whole line shows as selected.  However, when I copy/cut, I only copy/cut the right amount.  That is good, but it would be better if the highlight showed this.
image

linusha:
See https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace for more information.

The git conversion kicks in only when a file is actually committed. Therefore, I also enforced the line ending on system browser level. Doing this, is think I also discovered a bug in `updateSource`, which would never actually apply the source cleanups? 

Closes #187.

### DIFF
--- a/lively.ide/js/browser/index.js
+++ b/lively.ide/js/browser/index.js
@@ -815,10 +815,14 @@ export class BrowserModel extends ViewModel {
 
   updateSource (source, cursorPos) {
     const ed = this.ui.sourceEditor;
+
+    source = source.split(objectReplacementChar).join('');
+    source = source.replaceAll(/\r\n/g, '\n');
+
     if (ed.textString !== source) {
       ed.textString = source;
     }
-    source = source.split(objectReplacementChar).join('');
+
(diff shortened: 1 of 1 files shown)


merryman (approved):

merryman:
But we can still introduce bogus characters when using the lively text editor right?

linusha:
> But we can still introduce bogus characters when using the lively text editor right?

Yes. Do you think that is good or not? I was conflicted, since on one hand I think there should be a tool in lively that gives you the freedom:tm: to write dumb shit to disk, on the other hand this is, once more, a quite important fix, only working inside of the systems browser...

merryman:
Let's stick with the free as in freedom™️ 🕺approach and observe how this issue evolves.

PR #995, opened 2023-08-11
❄️: fix invalid `fontWeight` values leading to error on load attempt

linusha:
This should resolve this error

image

that occurred during the loading-screen. Only the default CSS string values are valid ('normal', 'bold') and numbers as Strings (we convert them internally).


merryman:
There are still some old `fontWeight: 'Medium'` references within the freezer components. (lively.freezer/src/loading-screen.cp.js, lively.freezer/src/ui.cp.js)

linusha:
I might be dumb but isn't that exactly the content of this PR? I might miss something obvious, but searching in my editor gives me this:
image

merryman:
Never mind, I screwed up.

merryman (approved):

PR #1004, opened 2023-08-17
Fix/freezer lean builds

merryman:
I noticed that lean production builds (excluding a lot of the core packages), were no longer running since:
1. We were insufficiently stubbing excluded modules in the build via the systemjs import map.
2. We had code in `lively.morphic/text` that was crashing when lively.ide was excluded from then bundle.

### DIFF
--- a/lively.freezer/src/bundler.js
+++ b/lively.freezer/src/bundler.js
@@ -65,12 +65,15 @@ const baseURL = typeof System !== 'undefined' ? System.baseURL : ensureFolder(pr
  * @param { object } warning - The warning object.
  * @param { function } warn - The custom error function to utilize for logging the warning.
  */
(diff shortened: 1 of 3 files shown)


linusha (approved):

linusha:
What is a lean production build? The standard build script with `--minified` set? Would be good to include that in the PR description before merging this.

merryman:
> What is a lean production build? The standard build script with `--minified` set? Would be good to include that in the PR description before merging this.

Good example for this is the following: https://github.com/engageLively/ibn-battuta-new/blob/main/tools/build.mjs
As you can see *a bunch* of packages are excluded here which constitutes what I would call a *lean* build with as little bundled code as possible.

PR #1005, opened 2023-08-20
Fix/path vertices reconciliation

merryman:
Fixes the reconciliation of the `vertices` property of Path and Polygon Morphs.

### DIFF
--- a/lively.ide/components/change-tracker.js
+++ b/lively.ide/components/change-tracker.js
@@ -97,6 +97,7 @@ export class ComponentChangeTracker {
     if (change.prop === 'position' && (change.target === this.trackedComponent || this.isPositionedByLayout(change.target))) return true;
     if (change.prop &&
         change.prop !== 'textAndAttributes' &&
+        change.prop !== 'vertices' &&
         !change.target.styleProperties.includes(change.prop)) return true;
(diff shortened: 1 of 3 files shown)


lively.ide/components/change-tracker.js
```diff
@@ -105,6 +106,7 @@ export class ComponentChangeTracker {
     if (change.selector === 'addMorphAt' && change.target.textAndAttributes?.includes(change.args[0])) return true;
     if (!change.selector &&
         change.prop !== 'layout' &&
+        change.prop !== 'vertices' &&
```
linusha, line 109: What are these two for? :thinking:

linusha (approved):

lively.ide/components/change-tracker.js
merryman: This case is for properties where the values are often updated in place. In those situtations `change.prevValue` and `change.value` reference exactly the same object and neither `obj.equals` nor `obj.deepEquals` helps us much. Right now I found this affects `layout` and `vertices`, however we may wanna manage this in a more canonical way in the future. We once had the policy in mind to *never* update properties in place, however that turned out to be not ideal for performance reasons, so we moved away from that idea.

PR #1012, opened 2023-08-22
Enhancement/list seperators

linusha:
This PR allows to customize some style properties of specific list items. Currently, `borderWidth` and `borderStyle` are supported, although one could generalize this approach.

This feature is used in the font family dropdown, to give an optical separation between the custom fonts for a project and the default lively fonts. 

image

### DIFF
--- a/lively.ide/studio/controls/text.cp.js
+++ b/lively.ide/studio/controls/text.cp.js
@@ -12,11 +12,11 @@ import {
 } from '../shared.cp.js';
 import { ColorInput } from '../../styling/color-picker.cp.js';
 import { PropertySection } from './section.cp.js';
-import { disconnect, connect } from 'lively.bindings';
+import { disconnect, epiConnect } from 'lively.bindings';
 
 import { DarkColorPicker } from '../dark-color-picker.cp.js';
 import { PaddingControlsDark } from './popups.cp.js';
-import { availableFonts } from 'lively.morphic/rendering/fonts.js';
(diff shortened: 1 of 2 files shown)


merryman:
The entry for the imported font in the list appears to be rendered twice. I have yet to find out if this is due to the separator logic or other causes. I will also push an adjustment that increases the padding between the separators and the entries.

merryman:
When I open a playground on this branch, the text control no longer lists the default fonts. I think it crashes since it assumes the `projectFonts` property always is present.

lively.ide/studio/controls/text.cp.js
```diff
@@ -116,16 +116,24 @@ export class RichTextControlModel extends ViewModel {
...
           return {
             value: font,
             string: font.name,
             isListItem: true
           };
-        });
-
-        fontFamilySelector.selection = text.fontFamily.replace(/^"(.*)"$/, '$1');
+        };
+        const projectFontItems = $world.openedProject.projectFonts.map(fontItemCreator);
```
merryman, line 129: This crashes when no projectFonts or openedProject is present.

merryman (changes requested):

lively.ide/studio/controls/text.cp.js
linusha: Should be resolved!

merryman (approved):

PR #1013, opened 2023-08-23
🎛️: rename video morph type -> codec

merryman:
Renames the `type` property in the video morph so it no longer conflicts with the `type` property we use in the `morph()` function calls.

### DIFF
--- a/lively.components/video.js
+++ b/lively.components/video.js
@@ -54,10 +54,10 @@ export class VideoMorph extends HTMLMorph {
           this.videoLoaded = false;
         }
       },
-      type: {
+      codec: {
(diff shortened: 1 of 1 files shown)


linusha (approved):
Alright from my side!
Do you have an eye out for code of @rickmcgeer that might break due to this? I think the `VideoMorph` was initially built for Ibn, but I am not sure.

merryman:
> Alright from my side! Do you have an eye out for code of @rickmcgeer that might break due to this? I think the `VideoMorph` was initially built for Ibn, but I am not sure.

Yes I do. In fact I was working right now on *that* particular code.

Issue #538, opened 2022-10-20
Font Management


linusha:
- The Font List in the properties panel is currently hard coded and may display several fonts that are not supported on the target system. **We should a) only display supported fonts and b) ship a wide collection of fonts to guarantee availability and compatibility**. This should work without Google Fonts.
- Remote fonts can be installed into the project via the Rich Text Control in the right side bar, where we allow to remote install new fonts from the font drop down. Remote fonts are fetched and loaded via an ensureFont() call that gets automatically updated inside the index.js of the project folder.
- Pay special attention that the font used on the landing page is bundled
- We only support woff files

PR #1015, opened 2023-08-23
🧰🎨: ignore underscore style props in reconciler

merryman:
We currently have several internal underscore style properties which have been introduced as part of #538. These incorrectly got reconciled upon component definition creation.


linusha (approved):
Nice. I like that we just exclude all _props. This is a good convention.
I will add it to the wiki as well.

PR #1018, opened 2023-08-24
🎨: fix order rendering of paths in css layouts

merryman:
Fixes the rendering of path/polygon morphs when positioned by css layouts.


linusha (approved):
image

For the sake of completeness, this is the problem this solves.

PR #1024, opened 2023-08-25
🎛️: do not retrieve image data if either height or width of canvas 0

merryman:
Fixes a crash in the renderer that would happen when the height of a canvas morph was set to `fill` in either height or width when controlled by a tiling layout.


linusha (approved):

linusha:
For further info see [here](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData?retiredLocale=de) under 'Exceptions'.

PR #1029, opened 2023-08-29
🎛️: make video cover entire video morph area

merryman:


lively.components/video.js
```diff
@@ -140,25 +140,6 @@ export class VideoMorph extends HTMLMorph {
     return this.videoDomElement ? pt(this.videoDomElement.width, this.videoDomElement.height) : null;
   }
 
-  onBoundsChanged (bounds) {
-    super.onBoundsChanged(bounds);
-    try {
-      if (!this.autorelayout) {
-      // once a user has resized, this turns off auto-resizing to accommodate the video
-      // the autorelayout flag is set by the resize methods to ensure that this
-      // won't happen when we are resizing to accommodate a video
```
linusha, line 149: There is quite a lot of code inside of this class that deals with resizing logic. I have to confess that most of it is a bit opaque to me. However, I am not sure it is a good idea to nuke this method and leave the rest in place? Do of all of the other flags still make sense, if we set width and height to 100% in the HTML forcefully?

At least the `autorelayout` flag I think is only read here, so this should go, as a minimum.

linusha (changes requested):

linusha (approved):

PR #1035, opened 2023-09-04
❄️🎨: compile available fonts statically into bundle

merryman:
Fixes the issue that within bundles `availableFonts()` would not return the correct result, since there is not project loaded inside the bundle. The solution is to statically compile the available fonts into the bundle.

### DIFF
--- a/lively.freezer/package.json
+++ b/lively.freezer/package.json
@@ -14,6 +14,7 @@
     "google-closure-compiler-osx": "^20200927.0.0",
     "wasm-brotli": "1.0.2",
     "rollup": "2.68.0",
+    "css": "3.0.0",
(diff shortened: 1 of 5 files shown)


linusha (approved):
LGTM. I like the approach of using the `css` library, this is much smoother than what I initially thought of, which would have been some regex magic (which was the reason for my fix being just throwing the check out of the window, as that seemed much more efficient :upside_down_face:).

I did not actually test drive this, but I assume you did so with e.g., the typeshift project?

merryman:
> I did not actually test drive this, but I assume you did so with e.g., the typeshift project?
Yes, I tried this with various projects. It appears to be working as intended.

PR #1041, opened 2023-09-27
Disable navigate back gesture in browser on swipe input

merryman:
When scrolling the world to the right via a touch input (trackpad on a laptop) this can trigger the back navigation gesture on some devices (at least mac OS). We not disable this feature by patching the document body inline style when the canvas mode is enabled on world load.


linusha (changes requested):
I'd think we need to update this when the mode gets toggled as well, an not only at load time. I pushed the change, please check and give me a thumbs up if you agree. Afterwards, I'll approve and merge.

merryman:
Looks good! I did not notice the toggle. Good catch!

linusha (approved):

PR #1042, opened 2023-09-27
Fix/reconciliation

### LINKED ISSUE #982: Reconciliation should auto-add and save necessary imports
**Describe the bug**
I reconciled a component and changed its fill to be a gradient. This necessitates the import of `LinearGradient` and `rect`. The correct core appears, but the calls are underlined and upon saving the module I am explicitly prompted to import them both.

**To Reproduce**
Open the `ThermometerConverter` from our talk demo and change the fill of one of the gauges.

**Version**: 3596b124e6edc271afc4baf4666b759442e25016

### LINKED ISSUE #990: Halos are offset for components with layouts and active `ViewModel`
**Describe the bug**
When designing a component, the halos are working fine. Activating the behavior of the component lead to the halo being offset. It might be related to the fact that the component that I edited is a prompt, although that is not certain.

**Screenshots**

https://github.com/LivelyKernel/lively.next/assets/14252419/a9b8da93-b7aa-436a-884d-02b352cf6152

**Version**: a1df97742c7848bebcabda2aa699b4de19b3dbcd

### LINKED ISSUE #1023: Several Issues with Reconciliation
Fix #982 :heavy_check_mark: 
 - [x] Imports that are inserted into a component module due to reconciliation are always absolute. It is better practice to turn them into relative imports in case they originate from the same project.
 - [x] There are issues with moving components around that lead to dropped `add()` calls or the incorrect creation of duplicate morphs (if they are moved down the morph hierarchy). :heavy_check_mark: 
 - [x] Live propagation among derived components and instances appears to be broken. I am pretty sure the `isDerivedFrom` check is failing for some reason. :heavy_check_mark: 
 - [x] reconciliation of effect properties is still broken. They sometimes insert `[object object]` strings into the code (cursor). :heavy_check_mark: 
 - [x] reconciliation of border removal via side bar does not do anything it seems. :heavy_check_mark: 
 - [x] reconciliation of chaning the height/width policy of text morphs inside the tiling layout is somewhat broken. It for instance does not reconcile the `fixedHeight` or `fixedWidth` property correctly when it is set to `hug` :heavy_check_mark: 
 - [ ] Structural changes in parent components should be reconciled in a more smart way. Right now nothing happens, which is maybe the worst of all worlds, leading to broken derived component definitions.

### LINKED ISSUE #1050: Clicked state can get stuck when mouse is moved away
**Describe the bug**
When clicking on a morph which has a designated master set for clicked, dragging the mouse of the morph and then lifting the mouse button, the morph will stay in the clicked state.

**Screenshots**
state_is_only_reset_when_hitting_bounds

**Version**: 0b0da66cca6e0f372e5af1a2cac287e9ccbd7d5d

### LINKED ISSUE #1057: Master States are not rest correctly when making the morph invisible fast
Peek 2023-10-09 15-56

When the morph gets made invisible while the state is still clicked, the state will not reset correctly, even after additional clicks.

### LINKED ISSUE #1064: Linter Warnings are missaligned when increasing the width of the System Browser
**Describe the bug**
When one increases the width of the System Browser, the linter warnings in the path status text will be missaligned. Clicking inside of the warning will fix this.

**Screenshots**

https://github.com/LivelyKernel/lively.next/assets/14252419/d6bb7050-ee0c-45a2-ace4-3ea409c91dca

**Version**: 6a89749513b773c0e01c023d221cc8a9cab7e931

### LINKED ISSUE #1093: Component Button in Browser does not reset when Component is removed with DEL instead of Halo
delete-editable-component

merryman:
Fixes #1093. :heavy_check_mark: 
Fixes #1023.
Fixes #1050. :heavy_check_mark: 
Fixes #1057. :heavy_check_mark: 
Fixes #1064. :heavy_check_mark: 
Fixes #990. :heavy_check_mark: 
Fixes #982. :heavy_check_mark: 

 - [x] Allow to override local default props. This can be achieved by not inserting a default value if there is an overriden master present that defines the prop explicitly.
 - [x] Come up with a better solution for temporarily hiding components when toggling behavior, such that the ghost halo effect (#990) can be avoided.

### DIFF
--- a/lively.components/window.js
+++ b/lively.components/window.js
@@ -739,9 +739,6 @@ export default class Window extends Morph {
 
   deactivate () {
     this.targetMorph.onWindowDeactivated();
-    // if (this.styleClasses.includes('inactive')) return;
-    // this.removeStyleClass('active');
-    // this.addStyleClass('inactive');
     if (this.master && this.master.auto === DefaultWindowInactive) return;
     this.master.whenApplied().then(() => {
       this.master.setState('inactive');
--- a/lively.components/window.js
+++ b/lively.components/window.js
@@ -5,10 +5,9 @@ import {
   morph,
   Morph
 } from 'lively.morphic';
-import { connect, once, signal } from 'lively.bindings';
+import { connect, signal } from 'lively.bindings';
 import { easings } from 'lively.morphic/rendering/animations.js';
-import { DefaultWindow, SystemWindow, DefaultWindowInactive } from './window.cp.js';
-import { PolicyApplicator } from 'lively.morphic/components/policy.js';
+import { SystemWindow, DefaultWindowInactive } from './window.cp.js';
 
 export default class Window extends Morph {
   static get properties () {
@@ -275,6 +274,7 @@ export default class Window extends Morph {
   buildHeader () {
     return morph({
       name: 'header',
+      acceptsDrops: false,
       extent: pt(this.width, 50),
       submorphs: [
         morph({
@@ -447,6 +447,7 @@ export default class Window extends Morph {
       submorphs: [
         rightResizer = morph({
           name: 'right resizer',
+          acceptsDrops: false,
           fill,
           width: resizerInset,
           draggable: true,
@@ -454,6 +455,7 @@ export default class Window extends Morph {
         }),
         leftResizer = morph({
           name: 'left resizer',
+          acceptsDrops: false,
           fill,
           width: resizerInset,
           draggable: true,
@@ -461,6 +463,7 @@ export default class Window extends Morph {
         }),
         bottomResizer = morph({
           name: 'bottom resizer',
+          acceptsDrops: false,
           draggable: true,
           fill,
           height: resizerInset,
@@ -468,6 +471,7 @@ export default class Window extends Morph {
         }),
         topResizer = morph({
           name: 'top resizer',
+          acceptsDrops: false,
           draggable: true,
           fill,
           height: resizerInset / 4,
@@ -475,6 +479,7 @@ export default class Window extends Morph {
         }),
         bottomLeftResizer = morph({
           name: 'bottom left resizer',
+          acceptsDrops: false,
           draggable: true,
           fill,
           extent: pt(resizerInset, resizerInset),
@@ -482,6 +487,7 @@ export default class Window extends Morph {
         }),
         bottomRightResizer = morph({
           name: 'bottom right resizer',
+          acceptsDrops: false,
           fill,
           extent: pt(resizerInset, resizerInset),
           draggable: true,
@@ -489,6 +495,7 @@ export default class Window extends Morph {
         }),
         topLeftResizer = morph({
           name: 'top left resizer',
+          acceptsDrops: false,
           draggable: true,
           fill,
           extent: pt(resizerInset, resizerInset),
@@ -496,6 +503,7 @@ export default class Window extends Morph {
         }),
         topRightResizer = morph({
           name: 'top rigth resizer',
+          acceptsDrops: false,
           draggable: true,
           fill,
           extent: pt(resizerInset, resizerInset),
--- a/lively.ide/components/editor.js
+++ b/lively.ide/components/editor.js
@@ -87,7 +87,8 @@ export class InteractiveComponentDescriptor extends ComponentDescriptor {
   }
 
   getASTNode (sourceCode = this.moduleSource) {
-    return getComponentNode(parse(sourceCode), this.componentName);
+    if (obj.isString(sourceCode)) sourceCode = parse(sourceCode);
+    return getComponentNode(sourceCode, this.componentName);
   }
 
   recordRemovedMorph (removedMorph, meta) {
--- a/lively.ide/components/editor.js
+++ b/lively.ide/components/editor.js
@@ -55,7 +55,7 @@ export class InteractiveComponentDescriptor extends ComponentDescriptor {
         prev.ensureComponentMorphUpToDate(c);
       }
       dependants.forEach(m => {
-        m.master._parent = newDescr.stylePolicy;
+        m.master = exprSerializer.deserializeExprObj(m.master.__serialize__());
       });
       newDescr.refreshDependants(dependants);
       prev.checkForGeneratedNames();
--- a/lively.ide/components/editor.js
+++ b/lively.ide/components/editor.js
@@ -262,5 +262,3 @@ export class InteractiveComponentDescriptor extends ComponentDescriptor {
     return descr;
   }
 }
-
-// component.DescriptorClass = InteractiveComponentDescriptor;
--- a/lively.ide/components/editor.js
+++ b/lively.ide/components/editor.js
@@ -1,4 +1,4 @@
-import { ComponentDescriptor, morph } from 'lively.morphic';
+import { ComponentDescriptor, morph, component } from 'lively.morphic';
 import { ExpressionSerializer } from 'lively.serializer2';
 import { string, obj } from 'lively.lang';
 import { module } from 'lively.modules/index.js';
@@ -153,8 +153,11 @@ export class InteractiveComponentDescriptor extends ComponentDescriptor {
    * Revert the component definition back to when we started the edit session.
    */
   async reset () {
+    const dependants = this.stylePolicy._dependants;
     await replaceComponentDefinition(this._backupComponentDef, this.componentName, this.moduleName);
     this._dirty = false;
+    this.stylePolicy._dependants = dependants;
+    // FIXME: carry over dependents for now
   }
 
   /**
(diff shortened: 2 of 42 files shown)


lively.components/window.js
```diff
@@ -7,7 +7,8 @@ import {
 } from 'lively.morphic';
 import { connect, once, signal } from 'lively.bindings';
 import { easings } from 'lively.morphic/rendering/animations.js';
-import { DefaultWindow, DefaultWindowInactive } from './window.cp.js';
+import { DefaultWindow, SystemWindow, DefaultWindowInactive } from './window.cp.js';
+import { PolicyApplicator } from 'lively.morphic/components/policy.js';
```
linusha, line 11:
```suggestion
```

lively.ide/studio/controls/section.cp.js
```diff
@@ -76,6 +76,7 @@ const PropertySectionActive = component({
 
 const PropertySectionInactive = component(PropertySectionActive, {
   name: 'property section inactive',
+  master: PropertySectionInactive,
```
linusha, line 79:
```suggestion
const PropertySectionInactive = component(PropertySectionActive, {
  name: 'property section inactive',
```
right?

lively.morphic/rendering/animations.js
```diff
@@ -125,11 +122,8 @@ export class PropertyAnimation {
       const shadowBefore = morph.dropShadow || new ShadowObject({ blur: 0, distance: 0, spread: 0 });
       const shadowAfter = config.dropShadow;
       config.customTween = fun.compose(p => {
-        morph.withMetaDo({ metaInteraction: true }, () => {
```
linusha, line 128: What is the rule for when we need `metaInteraction` to be `true` and when not? Below it has been added multiple times.

lively.morphic/tests/components-test.cp.js
```diff
@@ -767,30 +789,21 @@ describe('components', () => {
     expect(inst.get('bob').fill).to.eql(Color.red);
     expect(inst.get('alice').fill).to.eql(Color.blue);
     expect(inst.fill).to.eql(Color.cyan);
+    // this appears to be broken. The overridden master here
+    // needs to also overriden the masters in the sub policies.
+    // what needs to happen is a subPolicy.assignConfig(parent._autoMaster.getSubSpecFor(subSpecName)); // which will update the fields
     inst = part(c2, { name: 'troller', master: T1 });
-    expect(inst.master.overriddenMaster, 'correctly overrides top level master').to.eql(T1.stylePolicy);
+    expect(inst.master._autoMaster, 'correctly overrides top level master').to.eql(T1.stylePolicy);
```
linusha, line 796: :thinking:

lively.server/plugins/test-runner.js
```diff
@@ -9,7 +9,7 @@ export default class TestRunner {
     let results;
     try {
     this.headlessSession = new HeadlessSession();
-    await this.headlessSession.open('http://localhost:9011/worlds/load?name=__newWorld__&askForWorldName=false',  (sess) => sess.runEval(`typeof $world !== 'undefined' && $world.name == 'aLivelyWorld' && $world._uiInitialized`));
+    await this.headlessSession.open('http://localhost:9011/worlds/load?name=__newWorld__&askForWorldName=false',  (sess) => sess.runEval(`typeof $world !== 'undefined' && $world?.name == 'aLivelyWorld' && $world._uiInitialized`))
```
linusha, line 12: Is this an accident? In which cases will this kick in? We do have the `undefined` check before...

lively.ide/components/editor.js
```diff
@@ -144,17 +146,20 @@ export class InteractiveComponentDescriptor extends ComponentDescriptor {
...
   /**
    * Revert the component definition back to when we started the edit session.
    */
   async reset () {
-    await replaceComponentDefinition(this._backupComponentDef, this.componentName, this.moduleName);
+    const dependants = this.stylePolicy._dependants;
+    await replaceComponentDefinition(this._backupComponentDef, this.componentName, this.targetModule);
     this._dirty = false;
+    this.stylePolicy._dependants = dependants;
+    // FIXME: carry over dependents for now
```
linusha, line 162: Why is this necessary?

lively.ide/components/reconciliation.js
```diff
@@ -847,6 +926,11 @@ class MorphRemovalReconciliation extends Reconciliation {
         const [removedSpec] = exprBody.body.expression.elements;
         subExpr = subExpr.slice(removedSpec.start, removedSpec.end);
       }
+      if (subExpr.startsWith('add')) {
+        // extract the one element from the elements
```
linusha, line 930:
```suggestion
        // extract the removed element from the elements
```

lively.morphic/components/policy.js
```diff
@@ -572,28 +627,35 @@ export class StylePolicy {
...
-        if (node.master && node !== spec) {
-          return new klass({ ...obj.dissoc(node, ['master']), submorphs }, node.master, false);
-        }
+        if (node.isPolicy) return node.copy(); // duplicate the node to prevent in place modification or the original spec
         if (node.master) {
-          return new klass({ ...obj.dissoc(node, ['master']), submorphs }, node.master, false);
+          return new klass({ ...node, submorphs }, null);
+        }
+        // this is too weak, we need to check of there is any possible master (click, hover, auto, state, breakpoint) that
+        // may potentially provide the default value here
```
linusha, line 649: Does this still apply?

lively.ide/studio/controls/section.cp.js
merryman: right!

lively.morphic/tests/components-test.cp.js
merryman: No the comment stems from a time when this test was in fact broken. It no longer is, so it can go.

lively.morphic/rendering/animations.js
merryman: Originally this was a flag that was supposed to tell any tool that tracks/responds to changes, to ignore it. I think this has now been replaced by more `specific` flags, such as `skipReconciliation`. So we can certainly revise this flag, and remove/replace it throughout the system.

lively.server/plugins/test-runner.js
merryman: Good question, I did this in response to the test runner starting to crash on me. I did not investigate it further. Maybe it can go now?

lively.ide/components/editor.js
merryman: Dependents includes not only other components but also policies that are installed in morph instances. Unlike components, which are defined in modules and re-evaluate themselves, these instance policies can only be recorded at the point when they are initialized. So in order not to loose them, we need to "carry over" the dependents array.

linusha: I might be missing something obvious here, but I do not see how `replaceComponentDefinition` changes the `stylePolicy`? And this is the case where we would need to keep the `dependants` values around, no?

linusha:
Before we merge: 

- [ ] Please take a look at the remaining comments, especially the one asking if a comment does still apply.
- One Checkbox in one of the issues was not checked, namely:

> Structural changes in parent components should be reconciled in a more smart way. Right now nothing happens, which is maybe the worst of all worlds, leading to broken derived component definitions.

This seems to be actually unresolved as far as I can see, but we should merge this anyways.
- [ ] Let's file a new ticket for this one!
- [ ] We should file the `update` bug regarding the scene-graph as well.

linusha (commented):
I'll approve and merge once we have taken care of the above mentioned points. However, this looks good to me and I have checked the remaining issues we had before. I propose to code-freeze this, with the exception of the above points! :rocket:

lively.server/plugins/test-runner.js
linusha: I removed it and it works for me, I'll go ahead and change this back.

lively.ide/components/editor.js
merryman: `replaceComponentDefinition()` in this case reverts the souce code AND also reloads the module which causes the `stylePolicy` to be changed after it finishes execution. The `stylePolicy` before and after are different objects.

lively.morphic/components/policy.js
merryman: Well, right now this question is still not fully resolved. The way styles are calculated right now is that we check if there is a `autoMaster` and if no `autoMaster` is present (which is a fringe case to begin with) we proceed to utilize the default value of the Morph implementation. The comment raised the question if it makes sense to further dig for potentially more "appropriate" default values by taking a look at other masters such as the ones found in the breakpoints or the custom states. Since it is technically possible to construct policies that do not have an `autoMaster` (via the sidebar) I raised the question in the comment. I would be fine with adjusting the comment here, to make what I just explained more clear in the code and keep it just as a reminder for the future.

lively.ide/components/editor.js
linusha: Ah, got it!

lively.morphic/components/policy.js
linusha: Got it. I think it would be best if we extend the comment :slightly_smiling_face:

linusha (approved):

PR #1044, opened 2023-09-28
Add support for `RegExp` to `FileWatcher`

linusha:
This add the option to not only register a fileResource with a callback to the `FileWatcher` but, also a regex. Each time a file is saved, all actions that are registered for regexes that match the whole url of the file to be saved get executed.

### DIFF
--- a/lively.ide/world.js
+++ b/lively.ide/world.js
@@ -1563,18 +1563,14 @@ class FileWatcher {
   registerFileAction (fileResourceOrRegex, cB) {
     if (fileResourceOrRegex.isResource) {
       const file = fileResourceOrRegex;
-      if (this.fileActions[file]) {
-        if (Array.isArray(this.fileActions[file])) this.fileActions[file].push(cB);
-        else this.fileActions[file] = [this.fileActions[file], cB];
-      } else { this.fileActions[file] = cB; }
+      if (this.fileActions[file]) { this.fileActions[file].push(cB); } else this.fileActions[file] = [cB];
       return;
     }
+
     const newRegex = String(fileResourceOrRegex);
     const alreadyRegisteredRegex = this.regexToWatch.find((item) => item.regex === newRegex);
-    if (alreadyRegisteredRegex) {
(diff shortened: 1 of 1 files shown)


lively.ide/world.js
```diff
@@ -1546,32 +1546,48 @@ export class LivelyWorld extends World {
...
+    if (fileResourceOrRegex.isResource) {
+      const file = fileResourceOrRegex;
+      if (this.fileActions[file]) { this.fileActions[file].push(cB); } else this.fileActions[file] = [cB];
+      return;
+    }
+
+    const newRegex = String(fileResourceOrRegex);
+    const alreadyRegisteredRegex = this.regexToWatch.find((item) => item.regex === newRegex);
+    if (alreadyRegisteredRegex.actions) alreadyRegisteredRegex.actions.push(cB);
+    else alreadyRegisteredRegex.actions = [cB];
```
merryman, line 1569: Why not use a Set?

merryman (changes requested):

lively.ide/world.js
```diff
@@ -1545,33 +1545,47 @@ export class LivelyWorld extends World {
...
-    const actions = this.fileActions[savedFileData.resource];
-    if (Array.isArray(actions)) {
-      for (let action of actions) {
-        await action();
-      }
-    } else await actions();
+    const actions = this.filesToWatch[savedFileData.resource] || [];
+
+    for (const [k, v] of Object.entries(this.regexToWatch)) {
+      const regex = new RegExp(k);
```
merryman, line 1581: This will actually not work: `new RegExp(String(/.*.js/))` will not be equal to `/.*.js/`. Since stringification also includes the enclosing slashes the passed string is not the same regexp any more. Instead you need to rehydrate the regexp like so: 
```js
var parts = regexString.match(/\/(.*?)\/([a-z]*)/);
if (parts) {
  var pattern = parts[1];
  var flags = parts[2];
  var re = new RegExp(pattern, flags);
}
```

merryman (changes requested):
Does not work. See comment for details.

merryman (approved):

PR #1080, opened 2023-11-08
Update Tokens in remoteURLs upon loading of Project

linusha:
GitHub does only grant 10 tokens per oAuth-App/User combination. One gets a new token every time one logs in. This means, that after 10 login-logouts in lively, projects that still have an old token baked into their URL wont work anymore, as the token is revoked.
To mitigate this problem, we just always bake the currently logged in token into all remote urls once a project gets loaded. **Note, that this effectively makes a local lively install/locally running lively server a single user application at the moment, and logging in with multiple accounts could have unexpected consequences.**

### DIFF
--- a/lively.project/project.js
+++ b/lively.project/project.js
@@ -219,7 +219,17 @@ export class Project {
     loadedProject.url = url;
     loadedProject.configFile = resource(address.join('package.json').url);
     if (!onlyLoadNotOpen) {
+
       loadedProject.gitResource = await resource('git/' + await defaultDirectory()).join('..').join('local_projects').join(fullName).withRelativePartsResolved().asDirectory();
+      if (await loadedProject.gitResource.hasRemote()) {
+          const remoteURL = await loadedProject.gitResource.getRemote();
+          const userTokenInRemoteURL = remoteURL.match(/(gho_.*)@/)[1];
+          if (userTokenInRemoteURL !== currentUserToken()) {
+            const repoOwner = fullName.replace(/--.*/, '');
+            const name = fullName.replace(/.*--/, '');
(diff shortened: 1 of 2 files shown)


lively.project/project.js
```diff
@@ -218,7 +218,17 @@ export class Project {
     loadedProject.url = url;
     loadedProject.configFile = resource(address.join('package.json').url);
     if (!onlyLoadNotOpen) {
+
       loadedProject.gitResource = await resource('git/' + await defaultDirectory()).join('..').join('local_projects').join(fullName).withRelativePartsResolved().asDirectory();
+      if (await loadedProject.gitResource.hasRemote()) {
+          const remoteURL = await loadedProject.gitResource.getRemote();
+          const userTokenInRemoteURL = remoteURL.match(/(gho_.*)@/)[1];
```
merryman, line 225: This will crash once we introduce a `null` user token due to a non logged in user opening a project.

lively.project/project.js
```diff
@@ -218,7 +218,17 @@ export class Project {
...
+          if (userTokenInRemoteURL !== currentUserToken()) {
+            const repoOwner = fullName.replace(/--.*/, '');
+            const name = fullName.replace(/.*--/, '');
+            await loadedProject.gitResource.changeRemoteURLToUseCurrentToken(currentUserToken(), repoOwner, name);
```
merryman, line 229: This here is able to introduce a null user token if we are not logged in at all. This will make the project no longer loadable due to the line above.

merryman (changes requested):

merryman (approved):

PR #1090, opened 2023-11-13
📂: fix that `ProjectCreationPrompt` did not detect logged in user

linusha:
Previously, there was a bug where one could login on the landing-page, press the new project button and the creation prompt would ask for a new login, due to a race condition. In all other instances, the login would be detected correctly. This removes the race condition.


merryman:
I am seeing a cut off require when the new project prompt is shown:

weird-require

merryman (approved):

PR #1094, opened 2023-11-16
📂: fix layout jiggle on project prompt opening

merryman:
Fixes the strange wobble that happens on opening of the "new project prompt".


linusha (approved):
@merryman merging this is fine with me.
Two things that I observed:

1. I needed to run `make artifacts` to get the changes you made to kick in. There were no errors in the console. To my understanding, the rebuild should not have been necessary.
2. I have now force-pushed, but you might remember that the reconciliation inserted `name: undefined` into the code. Taking a closer look, the code that got inserted was even worse, something like this:
```js
part(InformIconOnLight,{name:undefined},{viewModel: ... });
```

PR #1095, opened 2023-11-17
Add Offline Mode

linusha:
This PR makes it possible to work with projects without an active internet connection. Right now, this needs to be actively enabled by the use, by toggling the offline mode either on the landing-page or in the top-bar.
Doing so will give the user some information about the restrictions in offline mode.

In offline mode, the following happens:

- When creating a new project, no github repo can be created.
- No projects can be created from a remote URL.
- Loading a project will not pull the corresponding repository.
- It is not possible to add a remote to an existing project.
- Saving a project will not push the project.
- Updated user data will not be available, but a logged in user stays logged in when toggling offline mode.
- Logging out the user while being in offline mode requires an additional confirmation, as a new login is not possible.

PR #1097, opened 2023-11-20
🧰: prevent all git operations when user is not logged in

### LINKED ISSUE #541: Deprecate the lively.user system against one that authenticates against GitHub
- [x] one should be able to log into lively with ones github account
- [x] this should enable a user to push/pull repositories
- [x] ~~not being logged in results in the same experience as being offline~~
- [x] the login can stay in the top right corner

# Research

- The POST endpoint of Github does not support CORS (known Issue), which means we either need to add a subserver or just use curl
- `curl -X POST -F 'client_id=d523a69022b9ef6be515' -F 'scope=repo' https://github.com/login/device/code` 
- curl -X POST -F 'client_id=d523a69022b9ef6be515' -F 'device_code=XXXX' -F 'grant_type=urn:ietf:params:oauth:grant-type:device_code' https://github.com/login/oauth/access_token
- With an access token we can use 
```
curl \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer XXXXX" \
  https://api.github.com/user
```
- to check whether we are logged in or not. We can use the `avatar_url` field of the response to display the github profile picture. 
- git clone https://XXX@github.com/user/repo.git -> might work with pull as well

linusha:
Initially, we planned without an offline mode for authenticated users. As we have that now, once #1095 is merged, our initial specification of not being logged becomes a bit nonsensical.

We now opted to prevent all creation/saving operations on projects once one is not logged in. Opening a project will still work and even pulling the latest changes, provided the currently stored token has sufficient permissions. The reason for this is basically to strongly incentivize users to log in, so that they do not accumulate uncommitted changes that increase the probability of nasty merge conflicts down the road. 

Closes #541.


merryman:
When I am a guest user on the landing page, then log in, then create a new project, the project creation fails with an error message "There was an error initializing the project or its remote.". However I am no longer a guest user but just still offline.

merryman (approved):

PR #1103, opened 2023-11-23
📂: exclude partsbin from project browser if not enabled explicitly

linusha:
As the partsbin should almost never be edited directly, this excludes it from being listes in the project browser. In case experienced users might want to access it this way regardless, one can set `livelyIncludePartsbinInList` in the `localstorage`. I opted for `localStorage` instead of `localconfig`, as that is not yet loaded on the landing-page.


merryman:
I see your point, however I am reluctant go down this route for the following reason: I believe "expert options" are fine, and can be hidden in local-storage (our entire module caching is), however they should always either be accessible via a UI element or via a piece of data that is separate from the client. So can we have a toggle in a menu somewhere that controls this local storage value?

merryman (approved):

PR #1118, opened 2023-11-28
Fix/tests

merryman:
When this is done, the test badge of lively.next will finally be green!


lively.ide/tests/js/import-helper-test.js
```diff
@@ -42,8 +43,12 @@ describe('import helper - injection command', function () {
 
   this.timeout(6000);
 
+  before(async () => {
+    await module('lively.morphic').revive();
+  });
+
```
linusha, line 49: Why do we need this explicitly here? It seems to be the only place where this is needed? :O

lively.ide/world.js
```diff
@@ -897,7 +897,8 @@ export class LivelyWorld extends World {
     if (target._responsiveHalo) target._responsiveHalo.remove();
     if (!Halo) return;
     if (!obj.isArray(target) && !target.halosEnabled) return;
-    if (this.halos().filter(halo => halo.target === target).length > 0) return;
```
linusha, line 900: I am confused - how were we ever able to show a new halo? :O

lively.lang/Path.js
```diff
@@ -66,7 +66,7 @@ Object.assign(Path.prototype, {
     // Does the Path resolve to a value when applied to `obj`?
     if (this.isRoot()) return true;
     const parent = this.get(obj, -1);
-    return parent && parent.hasOwnProperty(this._parts[this._parts.length - 1]);
+    return parent && Object.hasOwnProperty.bind(parent)(this._parts[this._parts.length - 1]);
```
linusha, line 69: Why is this necessary? This is hard to read :/

lively.modules/src/module.js
```diff
@@ -397,21 +397,27 @@ class ModuleInterface {
...
 
       // 2. update records of dependencies, so that they know about this module as an importer
       let impIndex; const hasImporter = dependencyRecord.importers.some((imp, i) => {
         if (!imp) return; impIndex = i; return imp && imp.name === this.id;
       });
-      if (!hasImporter) dependencyRecord.importers.push(record);
-      else if (record !== dependencyRecord.importers[impIndex]) { dependencyRecord.importers.splice(impIndex, 1, record); }
+
+      if (!hasImporter) record.depMap[dependencyRecord.name] = dependencyRecord.key;
+      // else if (record !== dependencyRecord.importers[impIndex]) { dependencyRecord.importers.splice(impIndex, 1, record); }
```
linusha, line 420: Huh?

lively.modules/tests/module-test.js
```diff
@@ -16,6 +16,10 @@ let S;
 describe('module loading', () => {
   beforeEach(async () => {
     S = getSystem('test', { baseURL: dir });
+    S.set('lively.transpiler', System.get('lively.transpiler'));
+    S.config({ transpiler: 'lively.transpiler' });
+    S.babelOptions = System.babelOptions;
+    S.translate = async (load) => await System.translate.bind(S)(load);
```
linusha, line 22: Why was this block not necessary before? Has nothing to do with this concrete test module, just wondering as these lines are inserted in a lot of tests now.

lively.morphic/tests/text/layout-test.js
```diff
@@ -7,7 +7,7 @@ const padding = Rectangle.inset(5);
 
 let w, h, t, tl, padl, padr, padt, padb; // eslint-disable-line no-unused-vars
 
-function text (string, props) {
+async function text (string, props) {
```
linusha, line 10: How does this make a difference? Nothing is awaited, no?

lively.ide/tests/js/import-helper-test.js
merryman: No, this covers up a bug in the freezer, that I forgot to fix.

lively.ide/world.js
merryman: Well, it only considered halos for a particular target? I assume this expression was `false` most of the time.

lively.lang/Path.js
merryman: If you create a javascript object like `Object.create(null)` and make it encounter this piece of code, it crashes. I have not exactly figured out why these kinds of object float around the system, but making the code more resilient like this, fixed the issues.

lively.modules/src/module.js
merryman: Importers is a synthesized property that does not make sense to transform in place. This is why the commented line does not make sense. It should be removed fully though.

lively.modules/tests/module-test.js
merryman: Since we migrated to the new version of SystemJS, there are a couple of things properties we further need to add to the System object for it to work fully. We can consider moving this setup code into the `getSystem()` itself, however that requires more testing and evaluating the impacts of this refactoring in the system.

lively.ide/world.js
linusha: Nevermind me, I read that as not greater than 0 :upside_down_face:

linusha (approved):

PR #1128, opened 2023-12-06
🎨🧰: properly handle component reviving and addition of unbundled changes

merryman:
Is a correction to #1105 where we apparently failed to properly revive component modules.

### DIFF
--- a/lively.ide/components/editor.js
+++ b/lively.ide/components/editor.js
@@ -35,8 +35,14 @@ export class InteractiveComponentDescriptor extends ComponentDescriptor {
   }
 
   static for (generatorFunction, meta, recorder, declaredName) {
-    const newDescr = super.for(generatorFunction, meta);
-    const prev = !recorder?.__revived__ && recorder?.[declaredName];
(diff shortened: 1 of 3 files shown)


lively.ide/components/editor.js
```diff
@@ -35,8 +35,14 @@ export class InteractiveComponentDescriptor extends ComponentDescriptor {
   }
 
   static for (generatorFunction, meta, recorder, declaredName) {
-    const newDescr = super.for(generatorFunction, meta);
-    const prev = !recorder?.__revived__ && recorder?.[declaredName];
+    const newDescr = super.for(generatorFunction, meta, recorder, declaredName); // force a new descriptor
+    if (recorder?.__revived__) {
+      // if we are in a bundle and this part of the bundle has been revived,
+      // we just implement the behavior of the base class. The interactive
+      // capabilities are not required in frozen parts of the system.
```
linusha, line 42: This is confusing to me - if we were revived, aren't we an interactive part, i.e., not a frozen one?

merryman: Yes, but by the time the module is revived the entire component is replaced by the version in the "live" module.

linusha (approved):

PR #1160, opened 2023-12-21
🧰: only trigger search for three or more characters

linusha:


lively.ide/text/search.js
```diff
@@ -418,6 +418,8 @@ export class SearchWidgetModel extends ViewModel {
         return;
       }
 
+      if (this.input.length < 3) return;
```
merryman, line 421: I do not think that this is the right place for this return.
This makes it impossible to search for any single character (searching for a single obscure character may be nessecary) or a variable name of 2 characters.

I would argue the purpose here is to avoid the unintended single or double character search that is invoked due to the typing into the search field. So I believe *blocking the search in response to input* is what we actually want to restrict.

merryman (changes requested):

lively.ide/text/search.js
linusha: You are of course right. Funnily enough, the better solution was right there all the time I think? I have now just increased to debounce timeout. For me the 500 worked reasonably well, what about you?

merryman: Works for me!

merryman (approved):

PR #1164, opened 2023-12-22
🎨: ensure custom tween changes are recorded in animations

merryman:
Animations that would use a `customTween` (both internally and by explicit parametrization) would not cause their changes to be registered as overrides in the style policies. This lead to funky behavior such as buttons disappearing (as can be witnessed in the inspector right now).


linusha (commented):
I can confirm that this works. However, I cannot really wrap my head around the changes you made and the description of the cause of the problem in the original comment. Could you elaborate for me?

Afterwards, I can approve and merge!

merryman:
I pushed a commit to the wrong branch. Will fix asap.

linusha (approved):

PR #1165, opened 2023-12-23
Fix/vertices reconciliation

merryman:
Fixes reconciliation and component definition creation of polygon/path morphs.
Also gets rid of a polygon animation glitch.

### DIFF
--- a/lively.morphic/morph.js
+++ b/lively.morphic/morph.js
@@ -3081,9 +3081,7 @@ export class PathPoint {
 
   moveBy (delta) {
     this.position = this.position.addPt(delta);
-    this.path.withMetaDo({ reconcileChanges: true }, () => {
-      this.path.onVertexChanged(this);
-    });
+    this.path.onVertexChanged(this);
     return this;
   }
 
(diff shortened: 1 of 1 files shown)


linusha:
What exactly is the scope of this PR? I can confirm that it is now possible to add Polygons into components and they are correctly reconciled once. What does not work for me is changing the vertices via the control points after the fact:


https://github.com/LivelyKernel/lively.next/assets/14252419/6d2d9171-202c-4c7a-ad40-fb554a278134

merryman:
> What exactly is the scope of this PR? I can confirm that it is now possible to add Polygons into components and they are correctly reconciled once. What does not work for me is changing the vertices via the control points after the fact:
> 
> 
> 
> 
> 
> https://github.com/LivelyKernel/lively.next/assets/14252419/6d2d9171-202c-4c7a-ad40-fb554a278134
> 
> 
> 
> 

You are correct, this PR should not have bricked the reconciliation in response to adjusting the vertices via direct manipulation. I will prepare a fix.

linusha (approved):

PR #1168, opened 2023-12-30
Fix/tests

merryman:
Fixes the currently broken tests.

### DIFF
--- a/lively.collab/comments/comment-count-badge.js
+++ b/lively.collab/comments/comment-count-badge.js
@@ -10,9 +10,7 @@ export class CommentCountBadgeModel extends ViewModel {
         },
         set (text) {
           this.ui.badgeLabel.textString = text;
-          if (this.morph) {
-            this.alignWithMorph();
-          }
+          this.alignWithMorph();
         }
       },
       morph: {},
@@ -32,7 +30,7 @@ export class CommentCountBadgeModel extends ViewModel {
   }
 
   alignWithMorph () {
-    this.view.position = this.morph.globalPosition.addXY(this.morph.width, 0).addXY(-5, 5);
+    if (this.morph) { this.view.position = this.morph.globalPosition.addXY(this.morph.width, 0).addXY(-5, 5); }
   }
 
   addToMorph (morph) {
--- a/lively.ide/js/objecteditor/index.js
+++ b/lively.ide/js/objecteditor/index.js
@@ -1058,7 +1058,7 @@ export class ObjectEditorModel extends ViewModel {
     }));
 
     let warnings = false;
-    [content, warnings] = lint(content, { 'no-unused-vars': ['warn', { args: 'none', varsIgnorePattern: `_|${className}` }] });
+    [content, warnings] = lint(content, { 'no-unused-vars': ['warn', { args: 'none', varsIgnorePattern: `${className}` }] });
 
     const editorSourceHash = string.hashCode(content);
     const { sourceChanged, outsideChangeWarning, selectedModuleId } = await this.withContextDo((ctx) => {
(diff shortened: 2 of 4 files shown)


linusha:
This hints at a bunch of stuff to think about, imo:

1. In CI, the test for the fixed proportional layout failed (see https://github.com/LivelyKernel/lively.next/actions/runs/7397667571). Locally, this was not the case for me. I adapted it now to match the one for the moving layout (which imo makes a lot more sense in general?) and now it seems to work both locally for me and in CI (https://github.com/LivelyKernel/lively.next/actions/runs/7398040068). The we have a situation where tests inconsistently fail between our local machines and CI sucks tremendously of course, however, as long as "only" graphical representation is concerned here I am willing to put this under "known issues"^^
2. The `startAndStopTicking` test is still flaky for me (just something I noticed while working on this).
3. The "http file listing contains properties of resource" test appears to be flaky (https://github.com/LivelyKernel/lively.next/actions/runs/7398040068/attempts/1).
4. I suspect that #1042 caused these, or do you have another culprit in mind? This once again raises the questions of running tests for each PR. We already discussed the idea of running the tests for all touched packages. As you rightly objected at that time, this will not catch all cases, as a lot of the core deps heavily rely on each other. However, as far as I can see, this would have helped us already in this case! Another option could be to only trigger the test running action once an approving review has been submitted, as a "last check" before merging.

Most of these discussions can be moved somewhere else. Please let me know if you agree with my changes to the tests and then I will merge this.

merryman:
> This hints at a bunch of stuff to think about, imo:
> 
>     1. In CI, the test for the fixed proportional layout failed (see https://github.com/LivelyKernel/lively.next/actions/runs/7397667571). Locally, this was not the case for me. I adapted it now to match the one for the moving layout (which imo makes a lot more sense in general?) and now it seems to work both locally for me and in CI (https://github.com/LivelyKernel/lively.next/actions/runs/7398040068). The we have a situation where tests inconsistently fail between our local machines and CI sucks tremendously of course, however, as long as "only" graphical representation is concerned here I am willing to put this under "known issues"^^
✅
>     2. The `startAndStopTicking` test is still flaky for me (just something I noticed while working on this).
Agree, but lets handle this in a separate PR.
>     3. The "http file listing contains properties of resource" test appears to be flaky (https://github.com/LivelyKernel/lively.next/actions/runs/7398040068/attempts/1).
Same as above.
>     4. I suspect that [Fix/reconciliation #1042](https://github.com/LivelyKernel/lively.next/pull/1042) caused these, or do you have another culprit in mind? This once again raises the questions of running tests for each PR. We already discussed the idea of running the tests for all touched packages. As you rightly objected at that time, this will not catch all cases, as a lot of the core deps heavily rely on each other. However, as far as I can see, this would have helped us already in this case! Another option could be to only trigger the test running action once an approving review has been submitted, as a "last check" before merging.
No, the culprit here was definitely #625.
> 
> Most of these discussions can be moved somewhere else. Please let me know if you agree with my changes to the tests and then I will merge this.

linusha:
> No, the culprit here was definitely #625.

Ah, this one I must have overlooked when going through the list of merged PRs! My point regarding the possible solutions for a more resilient CI check still stand though, I think. Thoughts @merryman?

linusha (approved):

merryman:
> > No, the culprit here was definitely #625.
> 
> Ah, this one I must have overlooked when going through the list of merged PRs! My point regarding the possible solutions for a more resilient CI check still stand though, I think. Thoughts @merryman?

Yes with regards to the basic "check affected packages test", I would vouch for giving it a try. Go ahead and prepare a action which does the job and lets see how it goes.

PR #1169, opened 2023-12-30
Fix/offline bundling support

merryman:


lively.freezer/src/resolvers/node.cjs
```diff
@@ -57,11 +58,17 @@ function ensureFileFormat(url) {
 
 // fixme: if we are bundling from a node.js script but targeting the browser, we need to properly resolve /npm- urls
```
linusha, line 59: This has nothing to do with the PR per se, but isn't this always the case, that we bundle from a node script? :thinking:

lively.freezer/src/resolvers/node.cjs
```diff
@@ -57,11 +58,17 @@ function ensureFileFormat(url) {
...
+  if (moduleName.startsWith('esm://cache/')) {
+    return moduleName;
+  }
   if (importer && importer.startsWith('https://jspm.dev/')) {
     if (moduleName.startsWith('/')) 'https://jspm.dev' + moduleName;
     if (moduleName.startsWith('https://jspm.dev')) return moduleName;
   }
+  if (importer && importer.startsWith('esm://cache/')) {
+    if (moduleName.startsWith('/')) 'esm://cache/' + moduleName;
+    if (moduleName.startsWith('esm://cache/')) return moduleName;
```
linusha, line 70: Why do we carry the slashes here? Above (for jspm) we do not do this, and that seems more logical to me? :thinking:

linusha (commented):
See above.

linusha (changes requested):
Also, this does not work for me? I checked out the branch, ran `make artifacts` and afterwards tried to bundle a project without internet connection. The bundling fails when SystemJS is tried to be received, which makes sense as we just load that from github? (https://raw.githubusercontent.com/systemjs/systemjs/0.21/dist/system.src.js)
This seems like something we might overthink in general, and just pull that one into the repository as well?

merryman:
> Also, this does not work for me? I checked out the branch, ran `make artifacts` and afterwards tried to bundle a project without internet connection. The bundling fails when SystemJS is tried to be received, which makes sense as we just load that from github? (https://raw.githubusercontent.com/systemjs/systemjs/0.21/dist/system.src.js)
> 
> This seems like something we might overthink in general, and just pull that one into the repository as well?

Yes, this is definitely something I overlooked.

lively.freezer/src/resolvers/node.cjs
merryman: This is an old comment, from the time when nodejs still would not support esm modules out of the box and would require bundling as well in order to be able to bootstrap its `lively.modules` runtime. We can basically remove this comment since it no longer applies.

lively.freezer/src/resolvers/node.cjs
merryman: True, fixed!

lively.freezer/src/resolvers/node.cjs
```diff
@@ -48,24 +46,29 @@ function isAlreadyResolved(url) {
...
-  if (moduleName.startsWith('esm://cache/')) return moduleName.replace('esm://cache/', 'https://jspm.dev/'); // for now
+  if (moduleName.startsWith('esm://cache/')) {
+    return moduleName;
+  }
   if (importer && importer.startsWith('https://jspm.dev/')) {
     if (moduleName.startsWith('/')) 'https://jspm.dev' + moduleName;
     if (moduleName.startsWith('https://jspm.dev')) return moduleName;
   }
+  if (importer && importer.startsWith('esm://cache/')) {
+    if (moduleName.startsWith('/')) 'esm://cache' + moduleName;
```
linusha, line 66: I somehow missed this looking at this code previously, but shouldn't this be returned or at least assigned somewhere? The same applies in line 62?

merryman: Yes, you are correct about this. Now with the changes of this branch, this code here is really no longer executed and can be removed, assuming we define all our esm imports via `'esm://cache/etc'` inside the code we bundle. This is because we keep resolving the esm imports to `esm_cache` urls, so the case that an importer is a `esm://cache` like id never appears.
Previously we would replace the `esm://` urls with `jspm://` analogs, leading us to ignore the cache entirely during bundling. That made the code above somewhat necessary.

linusha (approved):

PR #1176, opened 2024-01-09
🧰: stop open widget from being removed prematurely

linusha:
This fixes a nasty bug which lead to widgets being positioned at the top left corner of the world. The reason for this was that the binding which closes open widgets on scroll kicked in too early, which messed up the correct fading into the world of the widget. Delaying the setting of the open widget resolves this, as then the binding cannot kick in prematurely.


merryman (changes requested):
We need to apply this change to all of the controls, not just the point control. For example the number control also exhibits the same issue. As would the color popup if the property was scrolled into view.

linusha:
> We need to apply this change to all of the controls, not just the point control. For example the number control also exhibits the same issue. As would the color popup if the property was scrolled into view.

Yeah, naturally. That was :clown_face:.

merryman (approved):

PR #1167, opened 2023-12-30
Enhancement/landing loading page redesign

merryman:
The design of the world browser itself will be revised in the near future, when we will go about designing light/dark mode version of the entire lively.next UI.

closes #1156.

**This was estimated by both of us with 1w. This PR took 24h which is around 3d**

PR #1178, opened 2024-01-10
Enhancement/git repo check while loading

merryman:
This speeds up the loading of projects by kicking of the lengthy git pulls/version checks before the lively.next IDE is being fast loaded.
Can be reviewed once #1167 has been merged.

### DIFF
--- a/lively.project/project.js
+++ b/lively.project/project.js
@@ -270,6 +270,7 @@ export class Project {
           await loadedProject.gitResource.changeRemoteURLToUseCurrentToken(currUserToken, repoOwner, name);
         }
       }
+      // The latter part of the condition is only to not tightly couple lively.projects to the code in bootstrap.js
       if (!lively.isInOfflineMode && !lively.projectRepoPull) {
         // Ensure that we do not run into conflicts regarding the bound lively version.
         await Project.resetConfigFiles(loadedProject.gitResource);
--- a/lively.project/project.js
+++ b/lively.project/project.js
@@ -24,6 +24,8 @@ import { addOrChangeCSSDeclaration } from 'lively.morphic';
 import { generateFontFaceString } from 'lively.morphic/rendering/fonts.js';
 import { supportedImageFormats } from 'lively.ide/assets.js';
 import { evalOnServer } from 'lively.freezer/src/util/helpers.js';
+import { promise } from 'lively.lang';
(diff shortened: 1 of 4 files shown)


lively.project/project.js
```diff
@@ -236,16 +270,16 @@ export class Project {
           await loadedProject.gitResource.changeRemoteURLToUseCurrentToken(currUserToken, repoOwner, name);
         }
       }
-      if (!lively.isInOfflineMode) {
+      if (!lively.isInOfflineMode && !lively.projectRepoPull) {
```
linusha, line 273: When will this condition be true?

lively.project/project.js
```diff
@@ -33,10 +35,44 @@ export class Project {
     return localInterface.coreInterface;
   }
 
+  static async ensureGitResource (fullName) {
+    return resource('git/' + await defaultDirectory()).join('..').join('local_projects').join(fullName).withRelativePartsResolved().asDirectory();
+  }
+
+  static fetchInfoPreflight (fullName) {
+    let finishCheck;
+    ({ promise: lively.checkedProjectVersionRelation, resolve: finishCheck } = promise.deferred());
```
linusha, line 44: Just for my understanding - what happens here is that we store a promise inside of `lively.checkedProjectVersionRelation` so that we can do `await lively.checkedProjectVersionRelation` until the resolve callback is called, right?

The resolve callback in this case is called `finishCheck` and, despite that not really being transparent anywhere - that is a function and the `then` chain will also take care of calling it for us (despite the last entry being `finishCheck` and not `finishCheck()`) which will let anything waiting for `lively.checkedProjectVersionRelation` continue?

lively.project/project.js
merryman: Only if the prefetch did not happen. I did not want the projects to be dependent on bootstrap code in order to run on their own, so this is a safeguard for that.

merryman:
> This would have been drastically easier to review if the commits would have been separated better. Overall, this looks good, I'll probably merge right away after you took another look and addressed my comments.
> 
 
Yes, my brain sometimes tells me I should produce "runnable commits" which works only half of the time anyways and then I end up having the worst of all worlds.

Will prioritize semantic committing from now on.

linusha (approved):

PR #1180, opened 2024-01-10
🧩: append inserted import to first batch of imports in module

### LINKED ISSUE #728: import statement gets inserted at wrong position in code?
In `value-widgets.js`, there are import statements at the top and at the bottom of the file.
If this is desirable in general is imo debatable, but besides the point of this ticket.
A new auto-generated import got added at the bottom of the file, instead of the top where it would have belonged.

I assume what happens here is that we automatically insert new imports after the last import we find? We could solve this problem by automatically inserting the imports before the first import.

merryman:
Fixes #728.

**This issue was estimated to take 4h and took 1h.**

### DIFF
--- a/lively.modules/src/import-modification.js
+++ b/lively.modules/src/import-modification.js
@@ -82,8 +82,10 @@ export class ImportInjector {
       if (modified) return modified;
     }
 
-    // prepend new import
-    const lastImport = arr.last(imports);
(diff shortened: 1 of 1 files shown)


linusha (approved):
Checked the behavior with the exact scenario described in the original issue. Works.

PR #1184, opened 2024-01-10
Enhancement/ci tests for prs

linusha:
This PR introduces a CI check to make sure that our tests stay green. If a file in a lively core package is changed, we will run the tests of this package.

I checked for the following edge cases while developing this:

- a file that contains lively in its name is changed
- multiple files in the same package are changed
- we target a branch other than main with the PR
- a file inside of a subdirectory is changed

**Developing this took 5.5h, of which three were spent fighting with GitHub Actions :upside_down_face:.**

### DIFF
--- a/.github/workflows/check-deploy-docs.yml
+++ b/.github/workflows/check-deploy-docs.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout Repository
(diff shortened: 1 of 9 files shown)


linusha:
There is a problem when files in a subfolder are changed. I'll take care of it tomorrow.

merryman (approved):

PR #1185, opened 2024-01-10
Make bottom text of prompts easier to use

### LINKED ISSUE #778: Prompts should take their title separately from bottom text
Right now, prompts only take one text specification for their label. We achieve centered, bold titles by setting `textAndAttributes` and add a line-break.
However, this makes some text formatting combinations impossible.

For that reason, and also because it would make the usage of prompts way easier/nicer to read, we should add a separate Label and initialization option for the title and the bottom text of prompts.

linusha:
Closes #778.

**This was estimated to take 3 hours but took 10, due to me becoming derailed multiple times and hunting bugs that in the end did not exist.**


lively.components/prompts.cp.js
```diff
@@ -12,11 +12,18 @@ export class AbstractPromptModel extends ViewModel {
...
+        set (title) {
+          this.setProperty('title', title);
+          if (this.view) { this.ui.promptTitle.value = title; }
+        }
+      },
+      text: {
+        defaultValue: null,
+        set (text) {
+          this.setProperty('text', text);
+          if (this.view) this.ui.promptText.value = text;
```
merryman, line 26: The more canonical way to update in response to prop changes is to handle all of this inside the `onRefresh(changedPropName)` method and check for the prop name.

The idea here is to have a central space where all UI updates in response to property changes is handled instead of having them dispersed among a variety of setters inside the property definitions.

merryman (changes requested):

merryman (approved):

PR #1189, opened 2024-01-11
🧰: fix crash for dependents with master: false

merryman:
For components where `master: false` in order to invalidate inherited styles, the `InteractiveDescriptor.getDependents()` method would crash when component modules got saved.

### DIFF
--- a/lively.ide/components/editor.js
+++ b/lively.ide/components/editor.js
@@ -192,7 +192,7 @@ export class InteractiveComponentDescriptor extends ComponentDescriptor {
 
   getDependants (immediate = false) {
     return $world.withAllSubmorphsSelect(m =>
-      m.master?.uses(this.stylePolicy, immediate)
+      m.master?.uses?.(this.stylePolicy, immediate)
     );
   }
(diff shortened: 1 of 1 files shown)


lively.ide/components/editor.js
```diff
@@ -192,7 +192,7 @@ export class InteractiveComponentDescriptor extends ComponentDescriptor {
 
   getDependants (immediate = false) {
     return $world.withAllSubmorphsSelect(m =>
-      m.master?.uses(this.stylePolicy, immediate)
+      m.master?.uses?.(this.stylePolicy, immediate)
```
linusha, line 195: How comes that if `master: false` the first save-navigation operator does not suffice?

merryman: This was also news for me: apparently the conditional member expression only cancels further lookups when a requested member is `undefined`. Since `false` is not `undefined` the lookup proceeds and attempts to call the `uses()` function on `false` which fails.

linusha (approved):

PR #1190, opened 2024-01-11
Fix/browser dropping module controls

merryman:
Fixes the issue in the browser, where the clipboard controls would suddenly disappear forever when selecting a folder.

### DIFF
--- a/lively.morphic/rendering/animations.js
+++ b/lively.morphic/rendering/animations.js
@@ -109,6 +109,8 @@ export class PropertyAnimation {
   constructor (queue, morph, config) {
     this.queue = queue;
     this.morph = morph;
+    this.isStyleApplication = config.isStyleApplication;
+    delete config.isStyleApplication;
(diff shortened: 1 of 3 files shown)


lively.morphic/rendering/animations.js
```diff
@@ -109,6 +109,8 @@ export class PropertyAnimation {
   constructor (queue, morph, config) {
     this.queue = queue;
     this.morph = morph;
+    this.isStyleApplication = config.isStyleApplication;
+    delete config.isStyleApplication;
```
linusha, line 113: Why do we need to delete this?

linusha (approved):

lively.morphic/rendering/animations.js
merryman: Animations are messy and I wanted to avoid non morphing properties from entering the animations code. I will deal with the separately in a animations-refactoring issue. Maybe when we tackle the renderer altogether.

PR #1192, opened 2024-01-12
🚨🚨🚨 Fix for broken Github login flow

merryman:
This was an oversight from #1185. The condition that checked if `label` was an object would also return `true` if the label was an array since an array is also an object!!! Bug from hell.


linusha (approved):
I could have sworn I checked that. DIE MEISE man.

PR #1195, opened 2024-01-12
🎨❄️: improve `whenFontLoaded` performance

linusha:
If I remember our conversation about this correctly, the problem here was that we sometimes retrieved a `true` from `documents.fonts.check()` even though the font was not actually loaded yet, as the fonts CSS was not even loaded so far that the checked font was part of the `documents.fonts`.

Despite the comment I removed saying otherwise, this behavior is [not actually a bug](https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet/check#nonexistent_fonts) but consistent and logical behavior.

We spoke about using `check` only for fonts that are guaranteed to be part of the `FontFaceSet`. `FontFaceSet` provides a `has` method which I tried to use, but the `value` this method expects needs to be a `FontFace` object which we cannot just create on the fly based on a font name.

As we have all of the fonts that eventually become part of the `FontFaceSet` inside of the same CSS file, I opted to add an `onload` call to the loading of this file, which blocks calls to `check` as long as the CSS file is not yet fully loaded.

Again, I am not one hundred percent sure that this is what we talked about needing, but the result looked good to me.

### DIFF
--- a/lively.freezer/src/util/load-template.html
+++ b/lively.freezer/src/util/load-template.html
@@ -9,7 +9,7 @@
   <meta http-equiv="refresh" content="0;url=noscript.html">
   </noscript>
   <title>__TITLE_TAG__</title>
-  <link type="text/css" rel="stylesheet" id="lively-font-bundle" href="assets/fonts/fonts.css">
+  <link type="text/css" rel="stylesheet" id="lively-font-bundle" href="assets/fonts/fonts.css" onload="document.LIVELY_FONTS_LOADED = true">
   <link type="text/css" rel="stylesheet" id="lively-morphic-css" href="assets/morphic.css">
   <link type="text/css" rel="stylesheet" id="lively-project-css" href="assets/bundle.css">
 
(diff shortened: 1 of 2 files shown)


lively.freezer/src/util/load-template.html
```diff
@@ -9,7 +9,7 @@
   <meta http-equiv="refresh" content="0;url=noscript.html">
   </noscript>
   <title>__TITLE_TAG__</title>
-  <link type="text/css" rel="stylesheet" id="lively-font-bundle" href="assets/fonts/fonts.css">
+  <link type="text/css" rel="stylesheet" id="lively-font-bundle" href="assets/fonts/fonts.css" onload="document.LIVELY_FONTS_LOADED = true">
```
linusha, line 12: @merryman thinking about it we could also use the lively object here probably? I do not really care as this is something that should not be used anywhere else anyways 🤷

merryman: We can do that, I would just prefer moving the declaration of the `lively = {}` global in front of the link tag, to make sure we have no race condition, although I deem that unlikely tbh.

merryman (approved):

PR #1197, opened 2024-01-15
🛠️: rewrite artifact checker script to simply compare checksums

### LINKED ISSUE #699: Artifact Checker Script is unreliable
We recently had issues where the Artifact Checker script failed, probably because JSPM failed us.
However, this made me realize, that the artifact checker sometimes shows weird results. I could get it to fail consistently locally, even when it ran successfully remotely.
Also, we would probably want recursive dependency checking in the script.

Additionally, we should streamline the checking process altogether, since we still install/boot too often.

For testing purposes, [here's](https://github.com/LivelyKernel/lively.next/tree/artifact-checker-test) the relevant branch.

linusha:
Instead of the highly involced logic used previously, we just try to rebuild the artifact (if that already fails - bad news) and compare the MD5 checksum of the committed and new bundle.
Closes #699.

See that this finds how the installer cannot be bundled anymore here: https://github.com/LivelyKernel/lively.next/actions/runs/7532330435/job/20502689702?pr=1197.

### DIFF
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -16,6 +16,8 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: '20.10'
+      - name: Install `sultan`
+        run: pip3 install sultan 
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


linusha:
@merryman what is interesting: Initially, this failed on me in CI and not on my local system. Upon investigation I discovered that this was due to differences in the installed `base64-js` package versions. Locally, I only had one whereas in CI two were installed and the one I was missing was bundled inside of the `flatn` build in CI. Interesting is that just running `install.sh` on my local system did not solve this issue! This makes sense, as the lower version is introduced by `buffer` for which the version I had installed also suffices. This means that the actual number and version of packages installed by `flatn` is dependent on the order in which `flatn` installs everything. Just note worthy imo.

In the end I just deleted the package alltogether on my local system and ran `install.sh` which then resulted in both versions being present and now everything works as intended.

merryman (approved):

PR #1200, opened 2024-01-16
🧰: do not clear halo/property panel focus when interacting with tools

linusha:


lively.ide/world.js
```diff
@@ -205,8 +205,11 @@ export class LivelyWorld extends World {
     const activeWindow = this.activeWindow();
 
     if (activeWindow && target === this) activeWindow.deactivate();
-
-    if (!target.isFlap) this.handleHaloCycle(evt);
+    if (
+      this.morphsInWorld.includes(target) || // keep halo and properties panel focus around when interacting with tools
```
merryman, line 209: Maybe I am super confused but isnt the expression doing exactly the opposite of what the comment says?
When `target` is a tool, the expression returns false unlike true as the comment suggests.

merryman (changes requested):

lively.ide/world.js
merryman: many-halos

linusha:
@merryman yeah, the previous state of this PR was a bit of a mess, don't know how I got there. Now it should work as expected and make a lot more sense.

I am not super happy with the `setTimeout` I smuggled in here, but I couldn't think of an alternative solution with adequate size, given that this is only necessary for a very small scoped problem.

merryman (approved):
Looks good now!

PR #1206, opened 2024-01-17
🧰: fix misconfiguration of 'code editor' in `Inspector`

linusha:
@merryman please confirm that this works on you system as well before merging.


merryman (approved):
Works on my machine.

PR #1207, opened 2024-01-17
🧰: forcefully update font color of `BooleanControl`

linusha:
The boolean controls in the inspector do not update their font color anymore. This is because in the tree logic, we mess quite a bit with the renderer. This solves this problem, but is quite hacky - if it makes stuff worse, I do not know. @merryman what do you think?

### DIFF
--- a/lively.components/tree.js
+++ b/lively.components/tree.js
@@ -314,7 +314,7 @@ export class Tree extends Text {
             this.computeTreeAttributes(nodes),
             false, false);
           this.invalidateTextLayout(true, false);
-        } else if (this._lastSelectedIndex) {
+        } else if (this._lastSelectedIndex && this._lastSelectedIndex !== this.selectedIndex) {
           this.recoverOriginalLine(this._lastSelectedIndex - 1);
         }
         this.lastTreeData = this.treeData;


merryman:
I pushed an alternative change, that is a little less hacky and should work for all properties of the type where we toggle colors. Let me know what you think.

linusha (commented):
I can confirm that this works and I agree, that this is definitely the better fix!

However, I am bit taken aback by the commit message. You write:

> Previously we would recover the attributes of line even if the selected line had not been altered at all.

But isn't this backwards? I mean the situation was that we did in fact alter the line (from true to false) and **still** recovered the, now invalid, attributes.

merryman:
> @merryman I can confirm that this works and I agree, that this is definitely the better fix!
> 
> However, I am bit taken aback by the commit message. You write:
> 
> > Previously we would recover the attributes of line even if the selected line had not been altered at all.
> 
> But isn't this backwards? I mean the situation was that we did in fact alter the line (from true to false) and **still** recovered the, now invalid, attributes.

Fixed the wording in the commit. I used *line* whereas I should have used *line index*.

merryman (approved):

PR #1209, opened 2024-01-18
🧰: fix non animated window theme toggle in system browser

### LINKED ISSUE #667: `open browse snippet` is broken for server backend browsers
https://user-images.githubusercontent.com/14252419/215741395-fc0d802a-df37-4833-aac1-b81c2c8f6819.mp4

In the video, you can see that a) the generated code is not immediately executable for a browser that has its back end set to server and b) the browser gets initiated in a weird state. This might be partially due to me fixing the snippet in a wrong way. However, the browser should never be in this kind of "in between state".

merryman:
Fixes #667.
@linusha I somehow cannot reproduce the broken snippet from the video any more. Maybe it still happens on your machine? That was the part I assumed to be hard...

**This issue was estimated to take 2d but actually took 0.5h**

### DIFF
--- a/lively.ide/js/browser/index.js
+++ b/lively.ide/js/browser/index.js
@@ -985,6 +985,8 @@ export class BrowserModel extends ViewModel {
         fill: theme.background, duration, easing
       });
     } else {
+      sourceEditor.fill = theme.background;
+      columnView.fill = theme.background;
     }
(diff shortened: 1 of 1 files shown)


linusha (approved):
Hm, it works for me too, now. We take those I guess?! :O

PR #1210, opened 2024-01-18
🎨: fix accidental style policy override on poly init

### LINKED ISSUE #1205: Gradient Picker contains deformed polygons and shows misleading color
As can be seen in the video, there are two (**distinct**) problems with the gradient picker right now. 

First, the bubbles indicating color stops are deformed both in the picker itself and on the gradient. Clicking inside of the bubble resolves this problem.

Second, the color that is shown inside of the bubble afterwards is not correct. 

https://github.com/LivelyKernel/lively.next/assets/14252419/8a848405-6f9f-47b5-87ec-70dfe0b9d1ce

merryman:
Fixes #1205.

### DIFF
--- a/lively.morphic/morph.js
+++ b/lively.morphic/morph.js
@@ -3720,8 +3720,14 @@ export class Path extends Morph {
 }
 
 export class Polygon extends Path {
+  static get properties () {
+    return {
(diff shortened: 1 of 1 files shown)


linusha (approved):
I unlinked #1205, as the second problem described therein (misleading colors) still exists.

Issue #316, opened 2021-08-29
Generate source maps for transformed code


merryman:
Basically this is a `import sourceMap from 'https://jspm.dev/source-map'` and start from there by including proper calls to
`SourceMapGenerator` objects within the `lively.source-transform` predominantly.
This will have an impact on our compilation time and should be an optional thing, but I feel like this may be worth it by now.
Debugging from the browser would become quite a bit more accessible for many.

merryman:
There is a new case for this, especially since the introduction of fast load. #1038 
Since it now has become increasingly hard to debug the code when execution is spread across bundled code, it would be worthwhile to provide source maps, at least with regards to the fast-load bundles.
This would require us to ramp up the freezer, such that it also creates source maps alongside the source code transformations.

merryman:
Alongside enhancing the source transform to handle source maps, we should also get rid of babel and all of the dependencies alongside it, since modern browser basically support everything syntactically we require.

merryman:
Another interesting option is to make the custom source transformation operate around [magic-string](https://www.npmjs.com/package/magic-string). That way get can generate a source map on the fly as part of our general source transformations. That would only require an upgrade to the visitors to operate on a string instead.

linusha:
We should also keep [`recast`](https://github.com/benjamn/recast) in mind as an option, as it operates on ASTs directly, similar to our source transformations.

PR #1213, opened 2024-01-19
Enhancement/improve freezer logging

### LINKED ISSUE #508: Fix build script logging
Currently the logging of the build scripts is quite messy and looks like it crashes during build (even when it runs through successfult). We need to:
 - Introduce proper logging info that is formatted accordingly
 - Also make the script return the correct value upon completion.

merryman:
Resolves #508. I am still figuring out a setup to see if the solution that triggers the exit code on a crash is working.

### DIFF
--- a/flatn/tools/build-cjs.mjs
+++ b/flatn/tools/build-cjs.mjs
@@ -3,34 +3,41 @@ import { rollup } from 'rollup';
 import commonjs from '@rollup/plugin-commonjs';
 import { flatnResolve } from '../module-resolver.js';
 
-const bundle = await rollup({
-  input: './index.js',
-  plugins: [
-    {
-      resolveId: async (id, parentURL) => {
-        // directly use flatn to resolve this shit
-        if (id.startsWith('node:'))
-          return { id: id.replace('node:', ''), external: true };
-        try {
-          if (id.startsWith('lively.')) {
-            return await flatnResolve(id, parentURL);
+try {
+  
+  const bundle = await rollup({
+    input: './index.js',
+    plugins: [
+      {
+        resolveId: async (id, parentURL) => {
+          // directly use flatn to resolve this shit
+          if (id.startsWith('node:'))
+            return { id: id.replace('node:', ''), external: true };
+          try {
+            if (id.startsWith('lively.')) {
+              return await flatnResolve(id, parentURL);
+            }
+            if (!id.startsWith('.')) 
+              return await flatnResolve(id, parentURL);
+          } catch (err) {
+            return null; 
           }
-          if (!id.startsWith('.')) 
-            return await flatnResolve(id, parentURL);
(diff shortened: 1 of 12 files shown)


merryman:
@linusha As far as my investigation goes we cannot control the exit codes of node.js from within the lively rollup plugin, at least not in the version `2.70` we are using right now. That means we need to adjust our custom build scripts to wrap rollup inside a `try {} catch` block and manually exit the process with the appropriate exit code. Not elegant, but since we generate these scripts most of the time anyways I guess its fair? Another thing we could look into is to upgrade to newer versions of rollup which handles errors and warnings in a more centralized fashion, however that I would do in a separate ticket. I remember that being a little tricky though since a bunch of things have changed between version 2 and 4.

linusha (changes requested):
Overall, this looks goo to me.
I could confirm that the exit code works as expected.

Two questions/requests:

1. I still see the `browserslist`/`caniuse` output in the log, despite you removing that dependency. They are still somewhere in our dependency chain, resulting in 
```
rowserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db
```
Even deleting the node modules did not work, as `install.sh` will happily download them again.
Is this intended? Can we silence that similar to the async/await transpilation?

2. You introduced a verbose config option for the freezer plugin. Can we make that one easily accesible, at least in the bundling scripts for projects? One easy solution I can think of would be to just provide two `npm run` targets, like `npm run build` and `npm run build-verbose` or something. I am open to alternatives. What do you think?

linusha:
@merryman As this is part of our current sprint, please also add time tracking information.

merryman:
> Overall, this looks goo to me. I could confirm that the exit code works as expected.
> 
> Two questions/requests:
> 
>     1. I still see the `browserslist`/`caniuse` output in the log, despite you removing that dependency. They are still somewhere in our dependency chain, resulting in
> 
> 
> ```
> rowserslist: caniuse-lite is outdated. Please run:
> npx browserslist@latest --update-db
> ```
> 
> Even deleting the node modules did not work, as `install.sh` will happily download them again. Is this intended? Can we silence that similar to the async/await transpilation?

Yes, I think they are somehow imported via babel. I think we wont get rid of them unless we rid ourselves from babel, which will happen with #316. 
> 
>     2. You introduced a verbose config option for the freezer plugin. Can we make that one easily accesible, at least in the bundling scripts for projects? One easy solution I can think of would be to just provide two `npm run` targets, like `npm run build` and `npm run build-verbose` or something. I am open to alternatives. What do you think?

I would prefer a flag, but I can add that to the scripts, yes!

merryman:
> > I would prefer a flag, but I can add that to the scripts, yes!
> 
> Works for me! For future reference: We can then later on control this flag via #1223.

Added the verbose flag now, please review!

PR #1216, opened 2024-01-20
🧩: properly reconcile named imports in presence of namespace imports

### LINKED ISSUE #892: Adding import inserts invalid code
**Describe the bug**
Importing a specific entitiy via the add import dialogue will insert invalid code, when the origin module is already imported as a whole via (*).

**To Reproduce**
Open the index.js of the system browser and add the explicit import for PackageRegistry.

**Screenshots**

https://github.com/LivelyKernel/lively.next/assets/14252419/8ba9fe6a-ac9e-4f5a-bd21-41c35debde43

**Setup**
- Linux

**Version**: 8f1b1560ac2a4f6b1994b68966e5698f888ea336

merryman:
Resolves #892.


linusha (changes requested):
Hm, no malformed code gets generated anymore, but I find the behavior still a bit confusing. I get thrown into a loop with the `import` list dialogue:

https://github.com/LivelyKernel/lively.next/assets/14252419/87b4346a-c7d7-48d8-b8e3-f89669c04ca8


I would expect that either both options can be chosen or only the one I can actually use is shown.

merryman:
> Hm, no malformed code gets generated anymore, but I find the behavior still a bit confusing. I get thrown into a loop with the `import` list dialogue:
> import.mp4
> 
> I would expect that either both options can be chosen or only the one I can actually use is shown.

Fixed now, please confirm.

PR #1220, opened 2024-01-22
Fix styling and encoding selector of `ColorPicker`

linusha:
Fixes the broken encoding selector in the `ColorPicker` and unifies the appearance of the encoders.

@merryman please take a look at especially two changes:

1. The `DarkColorPicker` was assigned as its own master. I believe this was totally bonkers - am I missing something here?
2. I believe the `DarkColorPicker` showcases a bug in the component system. As you can see, I had to make a lot of changes (e.g., regarding the `fontSize`) in the dark variant as well as in the light variant. As the dark variant inherits from the light one, I believe this to be bug.

### DIFF
--- a/lively.ide/studio/dark-color-picker.cp.js
+++ b/lively.ide/studio/dark-color-picker.cp.js
@@ -1,9 +1,9 @@
-import { component, TilingLayout, part } from 'lively.morphic';
+import { component, TilingLayout } from 'lively.morphic';
 import { ColorPicker } from '../styling/color-picker.cp.js';
-import { Color, pt } from 'lively.graphics';
+import { Color } from 'lively.graphics';
 import { EnumSelector, TextInput, DarkNumberIconWidget, DarkThemeList, DarkCloseButton, DarkCloseButtonHovered } from './shared.cp.js';
 import { BackendButtonDefault } from '../js/browser/ui.cp.js';
-import { InputLineDark } from 'lively.components/inputs.cp.js';
+
 import { rect } from 'lively.graphics/geometry-2d.js';
 
 const DarkColorPicker = component(ColorPicker, {
--- a/lively.ide/studio/dark-color-picker.cp.js
+++ b/lively.ide/studio/dark-color-picker.cp.js
@@ -65,6 +65,7 @@ const DarkColorPicker = component(ColorPicker, {
                 borderRadius: 0,
                 submorphs: [{
                   name: 'value',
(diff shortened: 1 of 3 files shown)


merryman:
> Fixes the broken encoding selector in the `ColorPicker` and unifies the appearance of the encoders.
> 
> @merryman please take a look at especially two changes:
> 
>     1. The `DarkColorPicker` was assigned as its own master. I believe this was totally bonkers - am I missing something here?
Correct, I am sure this was due to a crazy reconciliation that happened at some point in the past.
>     2. I believe the `DarkColorPicker` showcases a bug in the component system. As you can see, I had to make a lot of changes (e.g., regarding the `fontSize`) in the dark variant as well as in the light variant. As the dark variant inherits from the light one, I believe this to be bug.
Unfortunately this is a feature. You are correct that defining the font size in the parent component should carry this value over to the derived instance. **HOWEVER** once you alter the style in a part of a derived component by assigning a master (which is done here by assigning the `DarkNumberWidget` to master) you reset the previously adjusted `fontSize` back to the one defined by `DarkNumberWidget`. The solution here would be to introduce a adjusted intermittent component, lets say `SmallDarkNumberWidget` that includes the adjusted smaller font. If you were to assign this instead of the original `DarkNumberWidget` you could save yourself the various repeated adjustments. In both cases by the way, since we can create both `SmallDarkNumberWiget` and `SmallNumberWidget`. You get the idea...

merryman (approved):
LGTM. Doesnt fix #1221 correct?

linusha:
> Unfortunately this is a feature. You are correct that defining the font size in the parent component should carry this value over to the derived instance. **HOWEVER** once you alter the style in a part of a derived component by assigning a master (which is done here by assigning the `DarkNumberWidget` to master) you reset the previously adjusted `fontSize` back to the one defined by `DarkNumberWidget`. The solution here would be to introduce a adjusted intermittent component, lets say `SmallDarkNumberWidget` that includes the adjusted smaller font. If you were to assign this instead of the original `DarkNumberWidget` you could save yourself the various repeated adjustments. In both cases by the way, since we can create both `SmallDarkNumberWiget` and `SmallNumberWidget`. You get the idea...

Ah, got it. I vaguely remember this, now that you told me again...

linusha:
> LGTM. Doesnt fix #1221 correct?

@merryman correct, does not fix that issue. In fact, I could only find out about that once I had the encoders working again :D

PR #1224, opened 2024-01-23
Fix interaction between loading indicators and commands

linusha:
This PR started with me wanting to introduce a loading indicator while the browser is loading/setting itself up (remember, we talked about this, @merryman). While at it, I discovered that commands already allowed a `progressIndicator` to be specified and in theory automatically display this. However, for asynchronous functions, this did not work as expected, as they are removed too early. Making the whole command handling asynchronous would be an extremely intrusive change in the system, so that option was off the table.
Instead, we can access the loading indicator from within the command function and just take care of there. To do this, I introduced a new property that a command can have  `handleLoadingIndicator`.
While I was working on that part of the system anyway, I fixed some commands where I observed broken behavior.

---

@merryman another thing I observed while working on this: Initially, I additionally tried to make the browser invisible while it set itself up and then only make it visible once everything is done. Doing so has **severe** implications on the text rendering, basically making the whole system unusable as the renderer fights for each line to appear after toggling the visibility on again. Care to reproduce this? If you can observer this as well, I will add it to the list of stuff I'll compile for you regarding a text rendering overhaul.

### DIFF
--- a/lively.ide/js/browser/ui.cp.js
+++ b/lively.ide/js/browser/ui.cp.js
@@ -1497,7 +1497,7 @@ const SystemBrowser = component({
 });
 
 async function browse (browseSpec = {}, browserOrProps = {}, optSystemInterface) {
-  const browser = browserOrProps.isBrowser ? browserOrProps : part(SystemBrowser);
+  const browser = browserOrProps.isBrowser ? browserOrProps : part(SystemBrowser, browserOrProps);
   if (!browser.world()) browser.openInWindow();
   browser.env.forceUpdate();
   delete browser.state.selectedModule;
--- a/lively.ide/world-commands.js
+++ b/lively.ide/world-commands.js
@@ -951,9 +951,9 @@ const commands = [
       let browser;
       if (args.reuse) {
         browser = browserForFile(System.decanonicalize(fileName)) || await Browser.open();
-        browser.browse(loc);
+        await browser.browse(loc);
       } else {
-        browser = await Browser.browse(loc, { extent: pt(700, 600) });
+        browser = await Browser.browse(loc);
       }
       browser.getWindow().activate();
       return browser;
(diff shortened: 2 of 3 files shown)


merryman:
> @merryman another thing I observed while working on this: Initially, I additionally tried to make the browser invisible while it set itself up and then only make it visible once everything is done. Doing so has **severe** implications on the text rendering, basically making the whole system unusable as the renderer fights for each line to appear after toggling the visibility on again. Care to reproduce this? If you can observer this as well, I will add it to the list of stuff I'll compile for you regarding a text rendering overhaul.

Isnt this problem just the fact that invisible text morphs currently have a broken `textLayout` since for this to work the morph needs to be rendered at some point? I guess the system browser just highlights this fact since it heavily relies on the editor's textLayout so this issue hits hard here.
But yeah, we should come up with some kind of a solution here in the future.

linusha:
> Isnt this problem just the fact that invisible text morphs currently have a broken `textLayout` since for this to work the morph needs to be rendered at some point? I guess the system browser just highlights this fact since it heavily relies on the editor's textLayout so this issue hits hard here.
> But yeah, we should come up with some kind of a solution here in the future.

Yes, that will be the driver of the problem. However, I think we can circumvent the necessity of already being renderer and actually think for "dumb" textmorphs we already do that. My first guess would be that the cause for this behavior is that the measuring takes place in the measuring layer of the morph itself, which causes problem when the whole thing is not rendered, as you pointed out. In those cases, we could try switching back to the global measuring layer. As there won't be a lot of manipulations, text changes, scrolls, etc. for invisible morphs, moving the dom nodes of the text could work here. For totally dynamic text that posed a problem, resulting in us utilizing the dedicated measuring layer more and more.

merryman (approved):

PR #1225, opened 2024-01-23
Fix Dropdown Tooltips

linusha:
The bug could easily be observed when looking at the tooltips in the fontfamily selector after scrolling down a bit. The problem here is that at the place where we set the tooltip, we already know that we are displaying a **new** item. Reusing the tooltip from `this` is thus bound to lead to invalid values, as we need to use `item` since both are guaranteed to be different items.


merryman (approved):

merryman:
Very good! This bug annoyed me for years.

PR #1227, opened 2024-01-24
Improve UX of right-clicks

linusha:
There are three things, all require some thoughts and testing, @merryman:

1. I preemptively leave the `onMouseDown` handler if the clicked button is  the right one. Please test if this works on your machine as well. I remember that we had something similar in the past, but ran into problems with regards to platform compatibility!? Another thing is: In theory, executing `onMouseDown` upon a right click is not wrong. However, I'd argue that we basically always use `onMouseDown` as 'left mouse button down'. What do you think? I encountered this problem because a lot of behavior can be toggled with the right mouse button, for example the side-bar flaps, which is unexpected imo.

2. This should be the least controversial one - just a fix for previously broken behavior of text that we never thought of when implementing the smart text with the dynamic document.

3. A change that prohibits context menus on most morphs. We briefly talked about this, and the solution I opted for now is quite brutal. However, I actually think it leaves all options one would usually want on the table. As stuff in windows etc. gets brutally excluded, I additionally introduced a flag `allowContextMenu`, that allows us to override this behavior on demand. Otherwise, this PR would basically kill off all custom tooling in windows, which is, at least in theory, a handy thing to have (see qinoq).

### DIFF
--- a/lively.morphic/events/EventDispatcher.js
+++ b/lively.morphic/events/EventDispatcher.js
@@ -614,6 +614,9 @@ export default class EventDispatcher {
     this.activations++;
 
     let err;
+
+    if (method === 'onMouseDown' && evt.domEvt.button === 2) evt.targetMorphs = evt.targetMorphs.filter(m => m.handleRightClickAsLeftClick);
+
     for (let j = evt.targetMorphs.length - 1; j >= 0; j--) {
       try {
         evt.targetMorphs[j][method](evt);
--- a/lively.morphic/events/EventDispatcher.js
+++ b/lively.morphic/events/EventDispatcher.js
@@ -1,4 +1,5 @@
 /* global System */
+/* eslint-disable no-use-before-define */
 import { arr, fun, Path, promise } from 'lively.lang';
 import { pt } from 'lively.graphics';
 import config from '../config.js';
@@ -86,7 +87,7 @@ const focusTargetingEvents = [
   'cut', 'copy', 'paste'
 ];
 
-const textOnlyEvents = [
+const textOnlyEvents = [ // eslint-disable-line no-unused-vars
   'input', 'compositionstart', 'compositionupdate', 'compositionend'
 ];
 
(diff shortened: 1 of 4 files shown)


linusha:
@merryman ah, now we can finally observe our new test safeguard working in the wild! Nice one. I'll take a look at the tests once I got you input on the proposed changes. I see no value in taking care of that as long as we are not sure if this is the way we want to tackle this problem in general.

merryman:
> There are three things, all require some thoughts and testing, @merryman:
> 
>     1. I preemptively leave the `onMouseDown` handler if the clicked button is  the right one. Please test if this works on your machine as well. I remember that we had something similar in the past, but ran into problems with regards to platform compatibility!? Another thing is: In theory, executing `onMouseDown` upon a right click is not wrong. However, I'd argue that we basically always use `onMouseDown` as 'left mouse button down'. What do you think? I encountered this problem because a lot of behavior can be toggled with the right mouse button, for example the side-bar flaps, which is unexpected imo.

While this will prevent the master application and setting of some meta information in the event state, it does not really do much: For one, morphs that implement their own `onMouseDown` still happily respond in the way the used to, regardless or not if super has an early return or not. Second, the bindings which are now used throughout the system still are getting invoked, and are unaffected by this logic.
If we were to consider `onMouseDown` just a *left button down* event then we need to restrict that within the event dispatcher, which I believe however to be too extreme. Then we either end up with not being able to respond to right clicks at all or having to introduce weird twin methods such as `onRightButtonMouseDown` and `onLeftButtonMouseUp` etc...

>     2. This should be the least controversial one - just a fix for previously broken behavior of text that we never thought of when implementing the smart text with the dynamic document.
Agreed!

>     3. A change that prohibits context menus on most morphs. We briefly talked about this, and the solution I opted for now is quite brutal. However, I actually think it leaves all options one would usually want on the table. As stuff in windows etc. gets brutally excluded, I additionally introduced a flag `allowContextMenu`, that allows us to override this behavior on demand. Otherwise, this PR would basically kill off all custom tooling in windows, which is, at least in theory, a handy thing to have (see qinoq).
Seems to be working for me so far. Regarding this flag, the name does not really lend itself that well. Its more of a forcing the context menu instead of allowing it right? Since the base case is that we allow it for morphs, but later filter out some if they are considered to be part of the IDE.

linusha:
> Regarding this flag, the name does not really lend itself that well

It is now named `forceContextMenu`.

---

Regarding the `onMouseDown` handling: As we discussed, I now opted for a solution directly inside of the event handler, that adjusts the `targetMorph` of the event if applicable. I also added a test for this.

merryman (approved):

PR #1230, opened 2024-01-24
🧰: fix pulling of `partsbin` repository

linusha:
Another case of "How did that ever work..." :disappointed:


merryman (approved):
Only worked as long as there were no other branches set as upstream in the partsbin repo I suppose ^^

Issue #174, opened 2021-02-05
Editor inserting bogus characters into text


rickmcgeer:
See the attached; some non-printing character is causing an error (or there's an editor bug)

![image](https://user-images.githubusercontent.com/656259/107073637-75921880-679c-11eb-886e-34040954f5dc.png)

And this is what happens when I try to Save
![image](https://user-images.githubusercontent.com/656259/107073765-9fe3d600-679c-11eb-9c1d-6c2e47d99b0a.png)

Copying to an editor buffer on my local machine shows no characters; copying and pasting back keeps the problem, and when I move the line the problem remains.

merryman:
What are the exact steps that that make these characters be inserted?

linusha:
I just got the same error message, but probably for another reason. I typed `!===` and upon saving I got that message. However, fixing my wrong syntax solved that.

merryman:
I am pretty sure this was due to line endings introduced on windows machines that caused trouble.
This was fixed in #940. @linusha Unless you object, I will go ahead to close this.

merryman:
I believe we still have not implemented a comprehensive mechanism that ensure these bogus newlines can not enter source files on a windows machine. That would need to be done for instance at the resource handle level. Right now we just do this on project saves and inside the system browser I believe.

linusha:
Fixed by #1239.

PR #1239, opened 2024-01-29
🪨: make sure to only write one type of line ending to disk

linusha:
This should solve #174 once and for all. Mind, that this of course makes it so that the actual writable contents inside of lively are now reduced. We talked about this and decided that people that actually require a specific line ending to be used for stuff to work are on their own. :clown_face:


linusha:
@merryman  I'll rebuild `flatn` once we have green-lighted this in principle.

merryman (approved):

PR #1240, opened 2024-01-29
🛠️: correctly measure the length of commit messages with emojis

linusha:
As graphemes are not correctly counted in most programming languages without using a librabry and Git🐙 bundles its own `node` version that hijacks the hooks so that we cannot use the new `Intl.Segmenter` API, this is a functioning workaround. The previous behavior became really annoying.

### DIFF
--- a/scripts/commit-msg-length-emojis.js
+++ b/scripts/commit-msg-length-emojis.js
@@ -1,0 +1,48 @@
+const emojiLengths = {
+    '🗨️' : 3 ,
+    '🌳' : 2,
+    '🎀' : 2,
+    '🔣' : 2,
+    '🛠️' : 3,
+    '🧑‍🏫' : 5,
+    '💭' : 2,
+    '🎛️' : 3,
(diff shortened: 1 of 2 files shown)


scripts/commit-msg-length-emojis.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+function emojifiedLength (string){
+    let len = string.length;
+    for (const emoji in emojiLengths) {
+        const re = new RegExp(emoji,'g');
+        const count = (string.match(re) || []).length;
+        if (string.includes(emoji)) len -= count * (emojiLengths[emoji]-1);
+    };  
+    console.log(len)
```
merryman, line 45: Looks a bit weird to have this function work by itself just via a `console.log`. Maybe just opt for a normal return here and put the `console.log` inside the `node -e` part of the git hook?

merryman (changes requested):

merryman (approved):

PR #1246, opened 2024-02-06
Branching Support for Projects

linusha:
It is finally here: Support for using multiple branches within projects!

> The best thing since the invention of sliced bread! - :stuffed_flatbread: 

---

This makes some assumptions about our intended workflow with `lively.next`. These are no new constraints, we already formulated them previously, but they have quite drastic consequences for this feature:

1. It is always better to give non-technical end-users something to work with, and let a programmer deal with the rest if necessary. We err on the side of "good enough" and just abort mission of more complex git operations fail. **We try to never actually loose work.**
2. It is assumed that one **saves** at the end of a lively session as soon as actual changes have been made. Not doing so drastically increases the chance that one lands in a situation where git knowledge is necessary to recover.

---

### Creating a new Branch

image

When saving a project, the dialogue will display the name of the currently checked out branch.
Expanding the "Advanced" section inside of the dialogue will give users the option to switch to a new branch and save the current changes on that one. **Currently, we do not provide another way of creating a new branch**. This makes it so that the trees that can be created with lively are always of the same type, arbitrary branching is not possible. However, to enable a PR focused workflow for designers and developers, this option suffices.
 
https://github.com/LivelyKernel/lively.next/assets/14252419/3d1ece27-9dba-43fb-b065-68315b3c77b6

### Opening Branches

On the landing-page, the Projects now also display a "Version" button, similar to that of worlds. It will open a list of all locally existing branches inside of the repository on disk as well as all branches existing on GitHub, in a merged fashion, so that branches that exist remotely and locally are not duplicated. Selecting a branch and then confirming by pressing the use button will use L2L and the shell to actually switch the branch inside of the repository, so that all subsequent operations will be performed on that branch without us needing to change the mechanism of how projects are loaded etc.

The list indicates whether a branch exists only locally, only remotely, or on both sides, indicate the currently checked out branch, as well as the default branch of the remote repository.

**We currently do not provide a way to delete branches, although adding that should be trivial**.

image

**In the case that we cannot easily migrate uncommitted changes from one branch to another when switching, we will show a warning in the console and keep the changes inside of a stash that indicate their origin. We will not prompt the user any further.**

### Cloning branches other than `main` when initializing a Project

When cloning a project from GitHub, we can select the branch to "land" in directly inside of the `ProjectCreationPrompt`.


merryman (changes requested):

merryman:
We also need a comprehensive message when the user is not logged in but attempts to retreive branch info. Currently the following happens:
stuck
errors

merryman:
I also pushed a minor fix to the styling of the branch entries in the version overview. Please confirm if you like them!

linusha:
> We also need a comprehensive message when the user is not logged in but attempts to retreive branch info.

Ah, good catch. I pushed something. As we allow the opening of projects without being logged in, we just retrieve the information available from the local repository, show a `StatusMessageError` that one should login to retrieve further information and highlight the user flap.

merryman (approved):

Issue #1223, opened 2024-01-23
Allow starting to bundle from within `lively.next` again (future of the Freezer Prompt)


linusha:
Right now, the Freezer Prompt is in a rather precarious state. As the bundling options for projects are now laid out in (generated) build scripts, we should bring it back.
A first checklist of what this endeavor would entail follows: 

The Prompt should at least be able to regenerate the buildscript with settings tweaked. We might also think about starting bundling processes via l2l from within lively.

It should be possible to configure
- the excluded modules
- minification?
- the new verbosity flag?

The Prompt itself requires some work, we still have some now unsupported layouts in there I believe and `bindings` targeting `viewModels` as well.

Here's a screenshot pointing out some of the things that needs work:

<img src="https://github.com/LivelyKernel/lively.next/assets/14252419/6012c76c-9ab2-46a9-9f38-0e37bde20e2b" width="350" height="500">

Additionally, the prompt invocation right now is not right. I suggest to only make the freezer prompt available via the button in the browser when one is inside of module with a valid `main` function. Right now, the button is e.g., visible when one browses `lively.morphic/index.js`.

Thoughts, @merryman?

rickmcgeer:
@merryman fixed the build script for  https://github.com/engageLively/galyleo-dashboard  (and thanks!) but the build script was put in the tools subdirectory of the project and only ran properly if it was invoked from the top-level directory.   Now, [the project] (https://github.com/engageLively/galyleo-dashboard) is probably a one-off, since:

1. We build several pages, not just one, from the project (studio-en, studio-jp, and publication)
2. The project has been developed simultaneously with lively.next, and so of course as design decisions in lively.next change it still bears the mark of old code.  

But still, I think it would be really useful to have a Freezer Guide which tells users how to freeze a project.

linusha:
@rickmcgeer Duly noted, we will work put this on our list of proper documentation to write!

PR #1258, opened 2024-02-14
🚨: small project fixes

linusha:
Fixes two problems with `lively.project`s:

1. Adapts the generated build scripts so that bundling is possible once more. @merryman uncovered a general problem: We currently generate this once (at creation time) and do not touch the script afterwards. This means that fixes like the one in this PR will not be automatically applied to already existing projects. We could now just force-create this with each save, thus guaranteeing that our fixes land in all projects. However, this will also override any customization one might have done to the bundling script. My proposed solution: We up the priority of #1223 and include therein thinking about a smart way of storing any customizations decoupled from the build script itself. Then, we can automatically generate an up to date and correct bundling script on each save. Thoughts?
 
2. Saving on the currently used branch was not possible due to an oversight when implementing #1246.

### DIFF
--- a/lively.project/prompts.cp.js
+++ b/lively.project/prompts.cp.js
@@ -504,7 +504,7 @@ class ProjectSavePrompt extends AbstractPromptModel {
         const branchName = branchInput.textString.trim();
         success = await this.project.gitResource.createAndCheckoutBranch(branchName);
         if (!success) $world.setStatusMessage(`Could not create branch ${branchName}, possibly due to a name collision!`, StatusMessageError);
-      }
+      } else success = true;
 
       if (success) success = await this.project.save({ increaseLevel, message, tag: this.tag });
       li.remove();
--- a/lively.project/templates/build.js
+++ b/lively.project/templates/build.js
@@ -34,7 +34,7 @@ const build = await rollup({
       ],
       resolver
     }),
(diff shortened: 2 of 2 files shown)


merryman (approved):

merryman:
> Fixes two problems with `lively.project`s:
> 
>     1. Adapts the generated build scripts so that bundling is possible once more. @merryman uncovered a general problem: We currently generate this once (at creation time) and do not touch the script afterwards. This means that fixes like the one in this PR will not be automatically applied to already existing projects. We could now just force-create this with each save, thus guaranteeing that our fixes land in all projects. However, this will also override any customization one might have done to the bundling script. My proposed solution: We up the priority of [Allow starting to bundle from within `lively.next` again (future of the Freezer Prompt) #1223](https://github.com/LivelyKernel/lively.next/issues/1223) and include therein thinking about a smart way of storing any customizations decoupled from the build script itself. Then, we can automatically generate an up to date and correct bundling script on each save. Thoughts?
> 
Yes, this is basically the good old migration problem we also faced with snapshots.
I would like to point out here, that currently a lot of unnecessary complexity is still exposed in the build scripts, that we can think about hiding entirely within the plugin. (The particular usage of the json plugin is one of those).
Once the interface for the rollup plugin itself is simplified, that could also simplify the complexity of the migration.

PR #1262, opened 2024-02-19
Dependencies in CI

### LINKED ISSUE #985: Projects with dependencies in CI
This is something that I discovered while working on, https://github.com/LivelyKernel/froscon-2023, specifically the demo deployment:

We spent a good amount of time thinking about ensuring correct resolving and retrieving of dependencies locally, but this logic is implemented entirely inside of the client.
This means, that trying to bundle some project with a dependency in CI will fail, as there is nothing ensuring the existence of the dependency there. This uncovers quite a large pile of stuff that we entirely missed until now: **They all resolve around the questions of how to retrieve the necessary dependencies, while in CI.**

# Running Tests of Projects with dependencies in CI

I think this is the easiest case. This is, because to run the tests, we need to load a lively world anyways, meaning the dependencies get downloaded just like locally. _This leaves open the question on how an error retrieving the dependencies manifests itself in the test pipeline, but I'd think this is secondary, at least for now._

**In case we go ahead and implement private projects, this becomes a bit more tricky, as we then need to have a method of authenticating to GitHub that we are allowed to download the dependency.** We however cannot use the current flow, as there is no human to enter the code. Github Actions allows some injection of token, I will need to look into this. We would therefore need a mechanism to use this token (an environment variable) inside of the client. This needs to be done in a way that does not impact the current workflow, centered around authenticating inside of the client.

# Building Projects with dependencies in CI

Again, we need to get the dependencies before the build can succeed. I identified three ways of achieving this:

1. The easiest option would be to open a client and load the project in there, essentially creating the same situation as above for the test running. However, this would force us to run the build processes of the lively next installation as well and in general, this solution would need a lot more time than necessary. I would prefer to do this as a last resort only.
2. Maybe this is the moment to fully embrace `flatn`? I think that could be cool, especially since we would end up with proper conflict resolution etc. but it seems overkill to me for now.
3. We could basically copy the resolution mechanism that ensures the existence of all dependencies in the client and put something similar in the bundler as well. Downside is, that the mechanism is quite complex, and we end up with a good amount of code duplication.
4. My preferred solution: We determine the whole dependency-chain at runtime in the client anyways. We also regenerate the pipeline-files at regular points. We could couple changing dependencies of a project to the recreation of the build file, and generate `git clone` steps for each dependency, ensuring their existence without the need to open the client. I think this is the easiest option to implement which also leaves us with the least amount of code in the end.

The credentials problematic remains the same as above, but since we can resolve the problem inside of the pipeline alone in this case, once we find out how to solve the problem inside of GitHub Actions, we are done and do not need to worry about somehow passing this inside of the lively client.

@merryman let's talk about this some time, please. I spent a good amount of time thinking about this, but want to make sure that I did not miss something obvious before proceeding.

linusha:
Closes #985.

This PR allows us to utilize the option of building/running tests in CI with projects that have dependencies.
The cases of public dependencies as well as private dependencies have been taken care of. In unusual circumstances (especially regarding multi-user usage of the same lively installation) it will be possible to break this. However, this scenario is rudimentary supported at best and really solving it would include a lot of security concerns which we decided not to tackle.

In general, the usage of private dependencies works by generating SSH keys in the browser and then using the private key as a secret inside of the project repository and uploading the public key to be used as a deploy key to GitHub. This way, the CI pipelines can authorize themselves to GitHub as being allowed to use the private repository. 

I also added a warning to the save project dialogue that gets shown when one adds a project as a dependency that is not available on GitHub, as that will usually inhibit other people from using/collaborating on that same project.

The setup can be seen working in [this repository](https://github.com/nextguys/project-demo).

**This was estimated to take 7.5d. Opening this PR took 5d.**

### DIFF
--- a/lively.git/index.js
+++ b/lively.git/index.js
@@ -1,1 +1,29 @@
-"format esm";
+/* global fetch */
+import { currentUserToken } from 'lively.user';
+export class GitHubAPIWrapper {
+  static async remoteRepoInfos (repoOwner, repoName) {
+    const token = currentUserToken();
+    const res = await fetch(`https://api.github.com/repos/${repoOwner}/${repoName}`, {
+      headers: {
+        accept: 'application/vnd.github+json',
+        authorization: `Bearer ${token}`,
+        'X-GitHub-Api-Version': '2022-11-28'
+      }
+    });
+    return await res.json();
+  }
+
+  static async listGithubBranches (repoOwner, repoName) {
+    const token = currentUserToken();
+    const res = await fetch(`https://api.github.com/repos/${repoOwner}/${repoName}/branches`, {
+      headers: {
+        accept: 'application/vnd.github+json',
+        authorization: `Bearer ${token}`,
+        'X-GitHub-Api-Version': '2022-11-28'
+      }
+    });
+    const branchData = await res.json();
+    // TODO: If this is called with a non-existing repository bad things happen!
+    return branchData.map(b => b.name);
+  }
+}
--- a/lively.git/index.js
+++ b/lively.git/index.js
@@ -86,6 +86,37 @@ export class GitHubAPIWrapper {
     });
   }
 
+  static async deleteRepositorySecret (repoOwner, repoName, secretName) {
+    const res = await fetch(`https://api.github.com/repos/${repoOwner}/${repoName}/actions/secrets/${secretName}`, {
+      method: 'DELETE',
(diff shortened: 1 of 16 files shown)


merryman (approved):
This looks fine, I guess this is one of those feature that we have to test for real in order to get a final verdict. The website can't come soon enough :-)

PR #1280, opened 2024-02-28
Fix/inspector issues

merryman:
Fixes various small issues with the inspector, each of which should be explained in the commits.

### DIFF
--- a/lively.ide/js/inspector/index.js
+++ b/lively.ide/js/inspector/index.js
@@ -282,7 +282,7 @@ export class PropertyControl extends DraggableTreeLabel {
   static renderEnumControl (args) {
     const { value, spec: { values }, keyString, target, node, tree } = args;
     const handler = async (evt) => {
-      const menu = target.world().openWorldMenu(evt, values.map(v => ({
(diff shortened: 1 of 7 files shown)


lively.ide/js/inspector/index.js
```diff
@@ -464,7 +465,7 @@ export class PropertyControl extends DraggableTreeLabel {
     const inspector = tree.owner;
     const handler = async (evt) => {
       // if already open, return
-      const editor = part(ShadowPopup, { hasFixedPosition: true });
+      const editor = part(ShadowPopupLight, { epiMorph: true, hasFixedPosition: true });
```
linusha, line 409: I might be missing somethin obvious here, but shouldn't we just set this to be true for all popups at all times? :thinking:

linusha (commented):
I proposed introducing this space to align all properties, even those without a drag indicator. Have you just forgotten or are you against this?

image

I do not find it to be super pleasing to look at either, but I'd prefer that over the misaligned properties. (The nicest looking solution would be to have the drag handlers aligned at the left side of the entries, but as we discussed this is not easily possible with the current implementation of the tree).

linusha (approved):

PR #1289, opened 2024-02-29
🛤️: perform secrets retrieval req via server as proxy

merryman:
Fixes an issue where a secrets retrieval request would get rejected by github due to CORS issues.

### DIFF
--- a/lively.git/index.js
+++ b/lively.git/index.js
@@ -1,6 +1,7 @@
 /* global fetch */
 import { currentUserToken } from 'lively.user';
 import { default as libsod } from 'esm://cache/libsodium-wrappers';
+import { resource } from 'lively.resources';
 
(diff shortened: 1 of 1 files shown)


linusha:
In principle I have nothing against this change. However, I am wondering how this came up? I was under the assumption that for the api subdomains this is not a problem. If this can become a problem, my next question would be why only for the secret retrieval?

merryman:
> In principle I have nothing against this change. However, I am wondering how this came up? I was under the assumption that for the api subdomains this is not a problem. If this can become a problem, my next question would be why only for the secret retrieval?

It came up when I tried to open up your personal branch on the lively.next website repo. But I openend other branches created by you in that past for different repos. So I can't really parse whats going on here. Are you sure this code path was executed at all times in the past?

linusha:
The reason seems to be that GitHub was not running smoothly.

PR #1292, opened 2024-03-01
📂: do not resolve save dialogue when pressing enter in description

linusha:


merryman:
Hitting enter now does not close the prompt any more, however I can't enter newlines anymore either.

merryman (approved):

PR #1339, opened 2024-03-14
Fix/reset derived components

### LINKED ISSUE #1336: Reverting a Reconciliation does not affect derived components
This is at least true for the name property.

### LINKED ISSUE #1337: Changing the name of a morph in a component does not get reconciled in resize policies of derived component

merryman:
Fixes #1337.
Fixes #1336.


linusha (changes requested):
We already talked about this, but for easier bookkeeping: When changing the name of a morph, derived components are correctly reconciled now. However, upon reverting this change using the blue button, the component itself will be correctly reverted but the derived component's source will stay the same and hence be broken.

linusha (approved):

PR #1357, opened 2024-03-25
🧰: fix alignment of component edit buttons on line wrap

### LINKED ISSUE #1349: Component Editing Buttons are laid out weird when line wraps
Screenshot from 2024-03-19 13-53-57

Wiggling the browser width leads to correctly laid out buttons most of the time. I am pretty sure this is an edge case with barely wrapping lines at the time of the initial display of a module.

merryman:
Resolves  #1349.
wrapping-component-edit-buttons

### DIFF
--- a/lively.ide/js/browser/index.js
+++ b/lively.ide/js/browser/index.js
@@ -1572,6 +1572,14 @@ export class BrowserModel extends ViewModel {
     }
   }
 
+  async isComponentEntity (entity) {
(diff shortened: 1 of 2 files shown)


linusha (changes requested):
Please make it so that the jump to a code entity via the muller columns targets higher lines, as the button might be hidden under the meta text box otherwise.

linusha (approved):

PR #1377, opened 2024-03-28
Fix Recursion when measuring static `Text`

linusha:
This resolves an bug that only became apparent once one begins to ensure, that the `textLayer` is always the same node. Due to the `extent` getter measuring the text and us accessing width/height we triggered a measurement in the measurement, which fucked up the state of the node that is moved around inside of the placeholder for the measurement.

Preventing this behavior here has no downside to the measuring process, as we only access the width/height when the dimension is marked as fixed in the properties, i.e. we can retrieve a value guaranteed to be correct without measuring.

### DIFF
--- a/lively.morphic/rendering/renderer.js
+++ b/lively.morphic/rendering/renderer.js
@@ -1939,6 +1939,9 @@ export default class Renderer {
    * @returns {Rectangle} The actual bounds of `morph` when rendered into the DOM.
    */
   measureStaticTextBoundsFor (morph) {
+
+    morph._measuringTextBox = true;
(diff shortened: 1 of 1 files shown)


merryman (changes requested):
I dont quite understand why the flag is set twice to the same value. Wouldnt it make more sense to set it to `false` in the end?

linusha:
> I dont quite understand why the flag is set twice to the same value. Wouldnt it make more sense to set it to `false` in the end?

:clown_face:

merryman (approved):

PR #1388, opened 2024-04-03
Open Browser in opened Project and restore Tab functionality

linusha:
Opens the browser in the directory of the currently opened project, if one exists. Also makes the tabs in the browser work again.


linusha:
@merryman as discussed, it would be extremely helpful if you'd investigate the bug that the x position of the tabs gets somehow set to 100.

linusha:
@merryman the eval backend button is still missing in action :(

merryman (approved):

PR #1418, opened 2024-04-17
Fix/system browser layout

merryman:
There were some things off with the layout structure of the system browser, which is fixed here.


linusha:
While the interaction in the current system is wild, the result is better than here: We cannot hide the columns completely right now, due to some heavily interacting layouts.
This changes that behavior:
image

Thus, we decided against merging this.

PR #1419, opened 2024-04-17
🧰: force layout in openend component morphs to ensure proper positioning

merryman:
Previously we would run into similar issues we already observed with the prompts, where large components whose ultimate extent is dictated by the layouts would get positioned completely incorrectly.


linusha (approved):
Just to make sure I understand the point here correctly: this is because we center the components as well, right?

PR #1420, opened 2024-04-18
Fix/system browser crash on save

merryman:
Fixes the funny issue where saving the system browser files *within* the system browser itself would lead to an error. 
Given this bug, this branch is a little extensive since it involved fixing various things about the way style policies where getting serialized.

### DIFF
--- a/lively.morphic/morph.js
+++ b/lively.morphic/morph.js
@@ -955,8 +955,7 @@ export class Morph {
         if (descr.readOnly ||
             descr.derived ||
             descr.isComponent ||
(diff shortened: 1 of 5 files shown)


linusha (approved):

lively.morphic/morph.js
```diff
@@ -955,8 +955,7 @@ export class Morph {
         if (descr.readOnly ||
             descr.derived ||
             descr.isComponent ||
-            (descr.hasOwnProperty('serialize') && !descr.serialize) ||
-            obj.equals(this[key], defaults[key])) {
```
linusha, line 959: This second line is now gone as well, was this intended?

PR #1432, opened 2024-04-22
Fix/world browser ld layout

merryman:
I was initially set out to fix some glitches in the world browser shown on the landing page. Ended up upgrading the constraint layout quite a bit.

### DIFF
--- a/lively.ide/studio/world-browser.cp.js
+++ b/lively.ide/studio/world-browser.cp.js
@@ -702,12 +702,14 @@ export class WorldBrowserModel extends ViewModel {
   }
 
   async displayItems () {
-    this.reset();
     const { loadingIndicator } = this.ui;
-    loadingIndicator.animate({
-      opacity: 1, duration: 300
-    });
+    if (!loadingIndicator.visible) {
+      await loadingIndicator.animate({
+        visible: true, duration: 300
+      });
+    }
     let entities = this.playgroundsMode ? await this.db.latestCommits('world') : await Project.listAvailableProjects();
+    this.reset();
     // Filter out the project that is opened anyways.
     if ($world && $world.openedProject) entities = entities.filter(availableProject => availableProject._projectName !== $world.openedProject.name);
     this.previews = entities.map(entity => {
@@ -732,10 +734,11 @@ export class WorldBrowserModel extends ViewModel {
 
       return placeholder;
     });
-
-    await loadingIndicator.animate({
-      opacity: 0, duration: 300
-    });
+    fun.debounceNamed('hide-ld-' + loadingIndicator.id, 300, () => {
+      loadingIndicator.animate({
+        visible: false, duration: 300
+      });
+    })();
     this.updateList();
   }
 
--- a/lively.ide/studio/world-browser.cp.js
+++ b/lively.ide/studio/world-browser.cp.js
@@ -642,7 +642,6 @@ export class WorldBrowserModel extends ViewModel {
   }
 
   viewDidLoad () {
-    this.ui.noWorldWarning.center = this.view.innerBounds().center();
     this.displayItems();
     this.onRefresh('showCloseButton');
   }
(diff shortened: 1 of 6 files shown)


linusha:
I a now missing the text when no result is found. The loading indicator is also not centered correctly.

https://github.com/LivelyKernel/lively.next/assets/14252419/e59bea93-a564-4771-90d8-d78164920bd3

lively.ide/studio/world-browser.cp.js
```diff
@@ -704,14 +702,15 @@ export class WorldBrowserModel extends ViewModel {
...
-      opacity: 1, duration: 300
-    });
+    if (!loadingIndicator.visible) {
+      await loadingIndicator.animate({
+        visible: true, duration: 300
+      });
+    }
     let entities = this.playgroundsMode ? await this.db.latestCommits('world') : await Project.listAvailableProjects();
     // Filter out the project that is opened anyways.
+    this.reset();
```
linusha, line 713:
```suggestion
    this.reset();
    // Filter out the project that is opened anyways.
```

Just to keep the comment attached to its line

lively.morphic/layout.js
```diff
@@ -1523,51 +1523,98 @@ export class ConstraintLayout extends Layout {
...
+        break;
     }
 
-    if (updateTransform) {
-      // trigger the halo if needed
-      morph.updateTransform();
+    return pt(width, height);
+  }
+
+  updateSubmorphViaDom (morph, node, makeDirty = false) {
```
linusha, line 1601: Maybe we should think about changing this name? As the whole purpose of this commit is to actually not read values from the DOM anymore, no?

merryman:
> I a now missing the text when no result is found. The loading indicator is also not centered correctly.
> missing_indicator_offset_loading.mp4

About the missing results message: That was actually supposed to only show up if there were no project *at all* to begin with. Hence the CTA that urged the user to create a new project. I fixed this, hence the message does not show up when one filters out everything from the available projects / worlds. I am open to adding an additional message separate for this case, which says something like 'no results found'.

Regarding the positioning, I can't seem to reproduce it. In my setup the loading indicator is perfectly centered. This sounds quite horrible, since it means that we have some kind of race condition at play, that depends on the speed of execution.
I guess the only way to find out is to debug this in a different setup. I will try.

linusha:
> About the missing results message: That was actually supposed to only show up if there were no project _at all_ to begin with. Hence the CTA that urged the user to create a new project. I fixed this, hence the message does not show up when one filters out everything from the available projects / worlds. I am open to adding an additional message separate for this case, which says something like 'no results found'.

Ehm I'm not sure about that. I can see that you changed the code that way, yes. I am pretty sure I implemented the previous behavior on purpose. The wording of the message makes sense in any case, as the message does not differentiate between no project existing at all or no project existing (matching the current filter). I do not care whether we add a different message, but I think there should be something when one has filtered out all existing items.

lively.morphic/layout.js
merryman: That is true, I refrained from refactoring since I did not want to overcomplicate the "DOM free" implementation of the layout. That will definitely come once we have all layouts transitioned (including grid).

merryman:
> > About the missing results message: That was actually supposed to only show up if there were no project _at all_ to begin with. Hence the CTA that urged the user to create a new project. I fixed this, hence the message does not show up when one filters out everything from the available projects / worlds. I am open to adding an additional message separate for this case, which says something like 'no results found'.
> 
> Ehm I'm not sure about that. I can see that you changed the code that way, yes. I am pretty sure I implemented the previous behavior on purpose. The wording of the message makes sense in any case, as the message does not differentiate between no project existing at all or no project existing (matching the current filter). I do not care whether we add a different message, but I think there should be something when one has filtered out all existing items.

Then I will revert to the old logic, and reword the cta a little.

lively.morphic/text/morph.js
```diff
@@ -2767,15 +2767,20 @@ export class Text extends Morph {
...
     if (
-      this._isDeserializing ||
-      !this.world() && !this.document ||
-      !this.visible ||
-      this.ownerChain().some(m => !m.visible) ||
+      !this.canBeMeasuredViaCanvas &&
+        (this._isDeserializing ||
+        !this.world() && !this.document ||
+        !this.visible ||
+        this.ownerChain().some(m => !m.visible)) ||
```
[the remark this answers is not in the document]
linusha: That makes sense, alrighty!

lively.morphic/layout.js
linusha: I'll open a ticket for that, so that we do not forget about it!

linusha (approved):

Issue #1430, opened 2024-04-22
Export to HTML raises an exception


rickmcgeer:
**Describe the bug**
Clicking on the morph menu (burger menu when morph is selected, top left corner of the halo) and then selecting `Export to HTML` gives this error:
```
Error: require is not defined
  Evaluating http://127.0.0.1:9011/dashboard/html-generator-6d86ead9.js
  Loading ./html-generator-6d86ead9.js
ReferenceError: require is not defined
    at Object.execute (http://127.0.0.1:9011/dashboard/html-generator-6d86ead9.js:21:13)
    at doEvaluateDeclarative (http://127.0.0.1:9011/dashboard/deps.js:3052:40)
    at ensureEvaluate (http://127.0.0.1:9011/dashboard/deps.js:2926:29)
    at http://127.0.0.1:9011/dashboard/deps.js:2551:16
```

**To Reproduce**
Select a morph on the sidebar, click on "Export to HTML" from the morph menu

**Screenshots**
![image](https://github.com/LivelyKernel/lively.next/assets/656259/bd66e0d2-25e8-49cd-89dc-c23d142a4cdb)

**Setup**

- Windows
- Linux

**Version**: Please paste the lively.next commit on which the problem occurred here (use the copy button of the Version Checker in the bottom left corner).
acaf0fc025a77f874d9c036b7827cdcaf624afee

merryman:
I believe this function is not really working any more, since we remove the virtual dom from the system. May I ask what you want to achieve with the exported HTML? I would honestly rather remove that functionality from the system than keep it.

rickmcgeer:
I'm just trying to do a freeze.  The scripts I used to do it from the command line are not working (see the issue I filed under the Galyleo Dashboard repo).  I have a bunch of features now working in the dashboard that I want to push, but the old freezer mechanisms don't seem to be there in the editor, and the scripts aren't working (bad includes).
So I'm sure there's a mechanism to do what I want (and I don't care if it's script or interactive), but I don't know what it is and Export to HTML seemed closest.

merryman:
`export to HTML` never did a freeze, all it did was literally copy the HTML code corresponding to a morph and export it to a file. It was really bare bones and barely useful to my knowledge. I believe @rksm used this functionality in the early stages of `lively.next` to build the current website.

Anyhow, since I now see that you rather accidentally encountered this functionality and it does not meet your needs, (even if it were to work), I would go ahed and remove that functionality from the system.

[...]

PR #1436, opened 2024-04-23
🧰: remove export to html menu item

merryman:
In response to #1430.


linusha:
Shouldn't we just remove all related code? :thinking:

linusha (approved):

PR #1471, opened 2024-05-03
Fix/text reconciliation

merryman:


linusha:
This fixes the following problems:

- insertions that are directly "duplicated" by the editor plugin fucked up the reconciliation before (for example ( which gets directly expanded to ())
- insertion of single or double quotes broke everything
- deletions of a part that should also trigger a deletion of a text attribute broke everything before (e.g. a single bold word that got completely deleted)

linusha (approved):

PR #1528, opened 2024-05-17
Fix/layout margin glitch

merryman:
Fixes this:
moving-weirdness

### DIFF
--- a/lively.morphic/layout.js
+++ b/lively.morphic/layout.js
@@ -899,8 +899,8 @@ export class TilingLayout extends Layout {
     const node = this.ensureLayoutComputed(container);
     const isPreliminary = node.getParent() && !node._computedMargin;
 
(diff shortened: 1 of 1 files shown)


linusha:
@merryman could you provide a screencast or reproducible example so that we can have so traceable content in the PR?

linusha (approved):
@merryman out of interest - why can we now remove all the `Math.floor` operations? Was this a previous attempt at tackling the same issue that is now obsolete as we know the real culprit?

merryman:
> @merryman out of interest - why can we now remove all the `Math.floor` operations? Was this a previous attempt at tackling the same issue that is now obsolete as we know the real culprit?

I believe this was stuff that I had around during the initial implementation of this layout but upon closer look, it appears not to be necessary. I think this was around the time when I was fighting with issues of border-width interfering with the internal computation of layouts and erroneously assumed it was due to rounding errors.

PR #1535, opened 2024-05-22
Fix/constraint layout loop

merryman:
Fixes an issue where constraint layouts would cause periodic render updates for no reason.


linusha:
@merryman IIRC, this was not the correct solution after all, as you found out after trying to explain what this actually does? I'll leave this as is for now.

linusha (approved):

Issue #1555, opened 2024-06-03
Project Creation Prompt is layouted wrongly


linusha:
![image](https://github.com/LivelyKernel/lively.next/assets/14252419/f8df6e94-e486-402e-8060-9824c72b1b54)

PR #1568, opened 2024-06-10
Fix/project prompt glitch

merryman:
Fixes: #1555


lively.morphic/text/morph.js
```diff
@@ -2349,6 +2349,7 @@ export class Text extends Morph {
     // [range1, attr1, range2, attr2, ...]
     // ranges are expected to be sorted and non-overlapping!!!
     this.document.setTextAttributesWithSortedRanges(textAttributesAndRanges);
+    delete this._allFontsLoaded;
```
linusha, line 2353: hmm

linusha (changes requested):
This messes with the measurements of annotations in menu items, presumably due to the changes to how text is measured that cannot respect the custom CSS style-class on canvas. The solution should be to exempt text that has a custom style class applied from the canvas measuring.

linusha (approved):

Issue #1560, opened 2024-06-03
Selecting and then "overwriting" text does not get reconciled correctly


linusha:
It is always necessary to explicitly hit delete and then write the new text in order for the reconciliation to work as expected.

https://github.com/LivelyKernel/lively.next/assets/14252419/13a7d973-d193-4dd8-9720-f68ddea745fe

PR #1569, opened 2024-06-11
🧰: resort to default patch if entire document replaced

merryman:
Fixes: #1560


linusha:
I have not tested this yet, but I do not think this suffices? The behavior described in the ticket is true even if one does not select the entire text in the document? Consider a textmorph which contains "Hello Robin", then select Robin and just write "Linus" instead, not hitting delete or replace before entering the new text.

merryman:
> I have not tested this yet, but I do not think this suffices? The behavior described in the ticket is true even if one does not select the entire text in the document? Consider a textmorph which contains "Hello Robin", then select Robin and just write "Linus" instead, not hitting delete or replace before entering the new text.

This is correct. Adjusted the fix to account for all replacement ops instead of just this special case.

linusha (approved):

Issue #1561, opened 2024-06-03
Tooling should prevent hugging Text from having text-wrap enabled


linusha:
It is quite easy to create textmorphs which have hugging (width) enabled and also try to break their content. The "dragged" default morph from the topbar has wrapping enabled, so just enabling hugging leads to such situations. However, those lead to weird looks and make editing quite painful:

https://github.com/LivelyKernel/lively.next/assets/14252419/155ec42c-1243-43ef-992c-97976aa91b69

PR #1600, opened 2024-07-03
🎨: always return line wrapping as "no wrap" if hugging enabled

merryman:
Simple yet effective way to deal with  #1561. What do you think @linusha?


linusha:
Let's merge this for now, but we should revisit this in the future, @merryman. I think #1621 is the right moment to do this, along with the improved color detection etc.

linusha (approved):

PR #1605, opened 2024-07-04
Fix/tiling layout issues

merryman:
Fixes a broken fill behavior and invalid margin computation in the tiling layout.

### DIFF
--- a/lively.morphic/components/policy.js
+++ b/lively.morphic/components/policy.js
@@ -275,8 +275,10 @@ export class BreakpointStore {
   }
 
   getLimitExtent ([v, h]) {
-    const hOffset = this._horizontalBreakpoints[h];
-    const vOffset = this._verticalBreakpoints[v];
+    let hOffset = this._horizontalBreakpoints[h];
+    let vOffset = this._verticalBreakpoints[v];
+    if (hOffset === 0) hOffset = this._horizontalBreakpoints[h + 1] - 1;
+    if (vOffset === 0) vOffset = this._verticalBreakpoints[v + 1] - 1;
     return pt(hOffset, vOffset);
   }
 }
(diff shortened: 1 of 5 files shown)


lively.morphic/layout.js
```diff
@@ -361,41 +363,7 @@ export class TilingLayout extends Layout {
...
-
-  resetYogaNodeFor (aMorph) {
-    if (!aMorph._yogaNode) return;
-    // clear and free the yoga object
-    aMorph._yogaNode.free();
-    delete aMorph._yogaNode;
-    // ensure that the morph gets rendered again, so that yoga is getting re-initialize
-    aMorph.makeDirty();
-    delete aMorph._lastComputed;
+    this.measureAfterRender(this.container); // no longer needed with yoga??
```
linusha, line 392: Can we either remove the code or the comment?

lively.morphic/components/policy.js
```diff
@@ -1052,7 +1052,7 @@ export class StylePolicy {
    */
   synthesizeSubSpec (submorphNameInPolicyContext, ownerOfScope, previousTarget) {
     const isRoot = !submorphNameInPolicyContext;
-    const transformProps = ['extent', 'position', 'rotation', 'scale', 'lineHeight'];
+    const transformProps = ['extent', 'position', 'scale', 'lineHeight'];
```
linusha, line 1057: "apply rotation for root of component for now" - why only 'for now'?

lively.morphic/layout.js
merryman: Remove, we will.

lively.morphic/components/policy.js
merryman: Because I am not yet sure if we run into unintended consequences. The reason I excluded transform properties from being applied to a morph when it is *the root* is that we mostly dont want that, even if these properties are not explicitly overridden.
For instance if we do not exclude the position from being applied to the root, it will lead to weird jump behaviors of morphs that suddenly align at positions that make no sense in the context they are in. For now I assume that rotation is seldomly defined at all, so it wont cause any issue. Also since it is applied so rarely you would assume that *if* a rotation is applied, it is there for a good reason.

linusha (approved):

PR #1628, opened 2024-07-23
🧰: add option to embed `HTML` in markdown

linusha:
Allows to use `HTML` code inside of `md` documents.


merryman (approved):

lively.ide/md/compiler.js
```diff
@@ -25,6 +25,7 @@ function addSourceLineMappingPlugin (md, options = {}) {
 class MarkdownCompiler {
   compileToHTML (src, options = {}) {
     let { linkedCSS, markdownWrapperTemplate } = options;
+    options.html = true; // allow usage of HTML tags in all markdown
```
merryman, line 28: We should also do the same in the parse method.

merryman (changes requested):

merryman (approved):

PR #1665, opened 2024-08-01
❄️: ignore dirs inside project assets during bundling

merryman:
Fixes a crash that happened if there were directories inside of the project asset directory.


linusha:
I assume this happened only when the folder were accessed by the `projectAsset` function, correct? And then this became problematic once you implemented the prerendering strategy?

We had a folder inside of the assets folder since forever, namely the `articles` folder for the blog. However, that one was never accessed by `projectAsset`.

merryman:
This is correct.

linusha (approved):

PR #1671, opened 2024-08-05
🧰: wrap lines by chars in terminal view for convenience

merryman:
For the longest time It was quite inconvenient to have to scroll sideways to see long lines in the terminal fully. This enables wrapping by default for the terminal in lively.


linusha (approved):
Ah yes, how annoying!

PR #1683, opened 2024-12-10
Fix/text performance

merryman:


linusha (approved):
After testing this, I believe we can do even better and remove the other attribute as well.
Please take a look and merge at your leisure. I tested this, but I am not 100% sure I thought of all consequences this has, as this logic of the textmorph is quite complex.

In any case, leaving the code that applies the now useless styleclass and the empty CSS class was sloppy, that can go for sure I'd believe?

PR #1689, opened 2024-12-19
❄️: wrap fast load toggle in separate container to fix alignment

### LINKED ISSUE #1655: `fastLoad` settings button is layouted wrongly on initial load
image

This fixes itself once one clicks on the button and subsequent loads in the same tab will preserve the correct position. But loading anew in a separate tab will again yield the result seen above.

merryman:
Fixed #1655

### DIFF
--- a/lively.freezer/src/landing-page.cp.js
+++ b/lively.freezer/src/landing-page.cp.js
@@ -271,7 +271,7 @@ class WorldAligningLandigPageUIElements extends ViewModel {
     $world._cachedWindowBounds = null;
     document.body.style.overflowY = 'hidden';
     this.ui.topSide.topRight = $world.visibleBounds().insetBy(10).topRight();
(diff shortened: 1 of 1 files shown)


linusha:
@merryman I can confirm that this works and will thus approve it.
However, I am not entirely clear as to why it works? Could you give me a short rundown either here or in person. I might would like to file this, as I do not think this solution is particularly pretty and I don't think I'd like to force this upon users of the system.

linusha (approved):

merryman:
@linusha The trick is to let the layouts handle the alignment instead of manual positioning logic. Here we are again faced with our (failed) attempt to make layouting logic synchronous in morphic. In this case the text layout of the toggle is not yet fully computed at the time we try to position it on load of the landing page. We then end up with this weirdly off fast load toggle described in the issue.

PR #1713, opened 2025-02-07
🧩: fix export lookup from frozen modules

merryman:
In fast load, when browsing for modules exporting certain entities, it would display all kinds of matches that would not make any sense. This fixes that.

### DIFF
--- a/lively.modules/src/export-lookup.js
+++ b/lively.modules/src/export-lookup.js
@@ -109,6 +109,21 @@ export default class ExportLookup {
     return exportsByModule;
   }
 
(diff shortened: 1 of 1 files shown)


linusha:
I might miss something obvious here, but could you remind me of how exactly I could verify that this works as intended?

merryman:
I think one easy example is if you open a workspace in fastLoad and select "Add Import" from the context menu, you will see a weird `__reexport___eval->` entry in the module prompt. That should be gone once you try the same on this branch.

PR #1719, opened 2025-02-19
Fix/disappearing drop

### LINKED ISSUE #1718: Dropping morphs into other morphs via halo fails to render morph.
**Describe the bug**
When dragging and dropping morphs with the halo, they disappear when dropped into a morph with one or more existing submorphs. This is not due to the underlying morph model but in fact a missing structural rendering change.

**Screenshots**
Image

**Setup**
- macOS

**Version**: e76487d620ffbb7002af43efaa6535cc4180708a (I am pretty sure this is due to ec17a0e6652db848868a824bd1ac1834120a2917)

merryman:
Fixes #1718.

Was due to an incomplete implementation of #1696 which was due to an oversight from my part: The full synthesization of the style properties of the entire morph hierarchy to be created from a component was faulty, since it ignores the effect of layout applications, which in the end leads to incorrect synthesization of style properties due to wrong breakpoints getting triggered.
This PR fixes that by performing a rough "pre-computation" of layouts on the level of style specs instead of morph, allowing us to avoid running into the issue described.

### DIFF
--- a/lively.morphic/components/policy.js
+++ b/lively.morphic/components/policy.js
@@ -969,14 +969,22 @@ export class StylePolicy {
 
   asFullySynthesizedSpec () {
     const extractBuildSpecs = (specOrPolicy, submorphs) => {
+      if (specOrPolicy.__alreadySynthesized__) return specOrPolicy;
       if (specOrPolicy.COMMAND === 'add') {
         specOrPolicy = specOrPolicy.props;
       }
       if (specOrPolicy.COMMAND === 'remove') return null; // target is already removed so just ignore the command
       if (specOrPolicy.isPolicy) return specOrPolicy.asFullySynthesizedSpec();
+      if (!submorphs) submorphs = specOrPolicy.submorphs || [];
       const modelClass = specOrPolicy.defaultViewModel || specOrPolicy.viewModelClass;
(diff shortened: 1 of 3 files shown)


lively.morphic/components/policy.js
```diff
@@ -1006,16 +1014,40 @@ export class StylePolicy {
         }
         synthesized.submorphs = transformedSubmorphs;
       }
+
+      if (synthesized.layout) {
+        synthesized.layout = synthesized.layout.copy();
+        synthesized.layout.estimateSubmorphExtents(synthesized, extractBuildSpecs); // should be done before the submorph specs are synthesized
+        // however for submorphs which themselves are resized by layouts, the extent information is not yet determined
+        // estimate the container extent
```
linusha, line 1022:
```suggestion
        // however for submorphs which themselves are resized by layouts, the extent information is not yet determined, instead estimate the container extent below
```

linusha (approved):

linusha:
For the sanity of our future selfs, I think it makes sense to fixup the comment I pointed out above. Afterwards, feel free to merge @merryman. I rebased it already for you, so that the conflict I introduced while merging is taken care of.

PR #1721, opened 2025-02-24
🎛️: fix labels of project browser on landing-page

linusha:
I assume this is due to the recent changes to the component instantiation. I am inclined to call this a bug - what do you think, @merryman? 
I think its okay to do it like this in this scenario, but if it turned out to be the case that we would now need to utilize `textAndAttributes` instead of `textString` in all/more cases, that would mean severely worse ergonomics from a user-perspective building their own components.


merryman:
I will look at maybe further refining the synthesization of text properties. The current solution only prioritizes between `textAndAttributes`, `value` and `textString` in an absolute manner, but I think we also need to take into account the current derivation level. I will prepare something that fixes that.

PR #1722, opened 2025-02-24
🎨: handle text props directly within the spec

merryman:
Alternative to #1721.

Previously, we would perform the normalization of text props at the very latest stage of synthesization, thereby loosing the overriding precedence of style properties for text properties.


linusha:
image
This does not solve the problem descibed in #1721.

merryman:
Sorry, should be fixed now.

linusha (approved):

PR #1745, opened 2025-03-08
Fix/docs pipeline

linusha:
It is not entirely clear to me why the lookup of past artifacts began to fail all of a sudden. I think this might be something that changed on the site of github - the delete-artifact action we used it the only non-official one, and we also needed to use quite an old version of that, so this might be the culprit.

I changed the logic of the action to only upload the build when triggering the workflow manually (i.e., when triggering a deploy of the updated docs). Previously, we uploaded that build on every run. Because of this, we introduced the explicit deletion, as there is a limit on the amount of artifacts we can upload. However, artifacts are retained for 90d only per default. We would need to deploy *often* to hit this now with the improved logic.

### DIFF
--- a/.github/workflows/check-deploy-docs.yml
+++ b/.github/workflows/check-deploy-docs.yml
@@ -46,7 +46,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Setup Pages
-        uses: actions/configure-pages@v3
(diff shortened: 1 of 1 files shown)


linusha:
@merryman merging this and afterwards rebasing the other affected PRs will take care of the failing action.

merryman (approved):

PR #1764, opened 2025-05-19
🧰: fix accidental overriding of wrong module on unsaved changes

### LINKED ISSUE #1753: Browser overwrites the wrong file
**Describe the bug**
I was editing studio/ui.cp.js in the Galyleo project, and added a function call to Dashboard in dashboard.cp.js.  I then clicked on dashboard.cp.js in the top browser pane, and an alert popped up asking me if I wanted to save changes.  I said yes, and then the browser _overwrote dashboard.cp.js with the contents of ui.cp.js_.
I was able to do a `git restore` in a shell to undo the damage, but the behavior was clearly a bug.
Moreover, even after I'd restored the file in the shell, the browser still had the wrong version of the file.  I closed all the browser windows, then opened a new browser, and it still had the bad version.  I had to quit Lively and restart it to get the right version of the file again.

**To Reproduce**
This should happen with any files in the browser.  Edit a file, click on another without writing, and then say save changes in response to the dialog.

**Screenshots**

**Setup**
Please delete non-applicable items!

- Windows
- Linux

**Version**: Please paste the lively.next commit on which the problem occurred here (use the copy button of the Version Checker in the bottom left corner).
c725a776b26bd744331765a4881ba4c0557d2fea

merryman:
Fixes #1753.

### DIFF
--- a/lively.ide/js/browser/index.js
+++ b/lively.ide/js/browser/index.js
@@ -1246,13 +1246,17 @@ export class BrowserModel extends ViewModel {
     return moduleLists.map(list => list.items.map(m => m.value)).flat();
   }
 
+  alreadySelectedModule (moduleUrl) {
+    return this.state.selectedModule?.url === moduleUrl && this.selectedModule.url === moduleUrl;
+  }
+
   async selectModuleNamed (mName, animated = true) {
     const columnView = this.ui.columnView;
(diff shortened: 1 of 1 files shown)


linusha (approved):
Note that I cannot really test-drive this, as I know of no reliable way to create the bug. Code looks promising though.

PR #1774, opened 2025-09-01
🤕: increase `protocolTimeout`

linusha:
Pretty strait-forward solution to the continuous timeout we have recently seen for the daily test runs.
120s might be higher than necessary, but for our use-case I honestly can't think of any downsides. Some breathing room won't hurt I guess...

@merryman, merging this, then rebasing #1773 should make your PR green as well - if not, this is not as good of a fix as I think. I ran multiple CI runs to check that the lively.lang tests succeed.


linusha:
Great, it didn't even work for this one :D

merryman:
I think this issue has been resolved after moving away from the old jspm servers.

PR #1794, opened 2026-04-17
Add lively.app: NW.js desktop distribution

merryman:
## Summary

- Add `lively.app/` sub-package that bundles lively.next as a standalone NW.js desktop app — users download one archive, double-click, lively runs. No clone, no terminal, no PATH setup.
- Multi-platform distribution (Linux x64, macOS arm64, Windows x64) built via a cross-platform Node build script (`lively.app/scripts/build.mjs`).
- Nightly GitHub Action (`.github/workflows/build-desktop-app.yml`) produces downloadable artifacts on the Actions run page.
- Drop google-closure-compiler deps (~571 MB) since SWC replaced them and the legacy path was gated behind an opt-in freezer UI flag.

## How it works

- **NW.js as the shell**: the desktop app is the NW.js binary + a bundled lively.next tree + a standalone Node.js binary for the server subprocess.
- **Server in-process where possible, child process where not**: ESM loader hooks (`module.register`, `registerHooks`, `NODE_OPTIONS --experimental-loader`) all crash NW.js's Blink renderer, so `start-server.cjs` (node-main) spawns `lively.server/bin/start-server.js` as a managed subprocess with flatn's `--experimental-loader`. A parent-PID watchdog (`watchdog.cjs`) preloaded into the subprocess ensures it dies when NW.js dies — including force-kill / crash — no orphans.
- **Dev mode vs bundled mode**: `start-server.cjs` auto-detects which mode it's in (based on whether `__dirname` is inside `rootDir`) and logs to `lively.app/boot.log` vs `~/.local/share/lively.next/boot.log` accordingly, sets up FLATN env itself, uses the bundled node binary when present.
- **install.sh hook**: after the freezer build, `setup.sh` fetches the NW.js SDK binary directly (the `nw` npm package's postinstall can't decompress through flatn's flat layout). Opt out with `./install.sh --no-desktop`.

## Bundle contents (1.9 GB uncompressed Linux)

- NW.js runtime (`nw` / `nwjs.app` / `nw.exe`, lib/, *.pak, ...)
- Standalone Node.js binary (~123 MB) for the server subprocess
- `app/` with the whole monorepo minus build artifacts, tests, docs, wrong-platform native bindings, puppeteer, and runtime caches
- `desktop/` with `start-server.cjs`, `watchdog.cjs`, `server-config.js`
- Platform launcher: `.desktop` + `launch.sh` (Linux), renamed `.app` (macOS), renamed `.exe` + `.bat` (Windows)

## Test plan

- [x] Clean-install the monorepo → produces working `./start-server.sh`
- [x] `node lively.app/scripts/build.mjs` → produces `dist/lively.next-linux-x64/`
- [x] Launch bundle with bare env (no nvm, no FLATN_*): `env -i HOME=$HOME PATH=/usr/bin:/bin DISPLAY=:1 bash dist/lively.next-linux-x64/launch.sh` → server boots in-bundle on 127.0.0.1, full lively.next dashboard renders (verified with screenshot)
- [x] Force-kill NW.js with SIGKILL → server subprocess exits within 1s (watchdog confirmed, no orphans)
- [x] `./install.sh --no-desktop` skips the NW.js SDK fetch
- [x] macOS arm64 bundle — first run via CI (build.mjs has branches, not locally verified)
- [x] Windows x64 bundle — first run via CI (build.mjs has branches, not locally verified)
- [x] Manually trigger `build-desktop-app.yml` on this branch before merge to smoke-test all three platforms

## Open follow-ups (not in this PR)

- The legacy `compileOnServer` / Terser+Closure code path in `lively.freezer/src/util/helpers.js` + the "Terser + Babel" option in `ui.cp.js` — dead code after the dep removal, clean up pass
- `lively.context` CDP hybrid (the original motivation for NW.js) — this PR is just the shell; the Smalltalk-style debugger integration using node:inspector is tracked separately in `plan_nwjs_app.md`
- Dev-deps pruning in `lively.next-node_modules/` to further shrink the bundle
- macOS code-signing + Windows code-signing for notarized / SmartScreen-friendly distributables

🤖 Generated with [Claude Code](https://claude.com/claude-code)


linusha:
@merryman have you tried what this means for lively.shell? Can the terminal escape from inside of the browser when it runs as a sub process in nw.js?

merryman:
> @merryman have you tried what this means for lively.shell? Can the terminal escape from inside of the browser when it runs as a sub process in nw.js?

Yes it still works the way it works on a local lively server setup. I configure it such that the shell cwd starts out in the user data directory of the app, where the user specific info such as the custom installed packages, custom projects, etc... are located.

PR #1799, opened 2026-05-14
🛠️: add Docker Compose development setup

tashabits:
## Summary

- Add a one-service Docker Compose development runtime based on Ubuntu 24.04.
- Install the host toolchain inside the image, including Node 24, Bun, Rust with `wasm32-wasip1`, Python tooling, browser libraries, fonts, and common CLI tools.
- Bind-mount the checkout into `/workspace` so source edits remain visible on the host, while Docker-generated directory state is isolated in named volumes.
- Seed default runtime files, run the existing install flow on first boot, and skip install on later boots when expected artifacts already exist.
- Harden freezer/browser boot paths found during end-to-end validation, including project-load diagnostics, browser-safe dependency handling, and linter loading.

## End-to-end Docker Compose guide

This path is optional. It is a local development launcher for hosts with Docker and Compose support installed, but without the native Node/Bun/Rust/Python/browser dependency stack. It does not replace the native install and run-server workflow.

1. Check out this branch and start from the repository root.

```bash
git checkout docker-compose-setup
docker compose up --build
```

2. Wait for the first-run install to finish. The first boot can take several minutes because the container runs the repository install flow and builds generated artifacts.

3. Open the running system from the host.

```text
http://localhost:9011/dashboard/
```

The service binds only to `127.0.0.1:9011`, so it is meant for local development rather than public deployment.

4. Work against the bind-mounted checkout.

```bash
docker compose exec lively bash
cd /workspace
```

Source edits made from inside the container are edits to the host checkout.

5. Treat Docker state as separate from native state.

Dependency, database, cache, and build directory contents are stored in Docker named volumes layered over the generated paths inside `/workspace`. Docker may leave empty mount-point directories in the host checkout, but the Linux container artifacts inside those paths are not shared with native installs.

Later container starts reuse those named volumes and skip install when the expected dependency tree, class runtime, and freezer bundle are present.

6. Stop the stack without deleting Docker-generated state.

```bash
docker compose down
```

7. Remove Docker-generated state and force the next Compose start to behave like a fresh Docker install.

```bash
docker compose down -v
```

For a truly fresh validation run, use a fresh clone/worktree of the branch and run `docker compose up --build` there. That avoids mixing existing generated artifacts with the first-run path being tested.

## Debugging and boot checks

The container includes the runtime dependencies needed to run the browser boot proof from inside the service. Because this repository uses the flat dependency layout, run the proof through the project environment setup.

Dashboard check:

```bash
docker compose exec -T lively bash -lc 'cd /workspace && . scripts/lively-next-env.sh && lively_next_env /workspace && node --no-warnings --experimental-import-meta-resolve --experimental-loader ./flatn/resolver.mjs ./scripts/check_boot.js http://127.0.0.1:9011/dashboard/ 60000'
```

World-load and edit proof:

```bash
docker compose exec -T lively bash -lc 'cd /workspace && . scripts/lively-next-env.sh && lively_next_env /workspace && url="http://127.0.0.1:9011/worlds/load?name=__newWorld__&askForWorldName=false&fastLoad=true" && node --no-warnings --experimental-import-meta-resolve --experimental-loader ./flatn/resolver.mjs ./scripts/check_boot.js "$url" 120000 --require-world --prove-edit'
```

The boot proof prints structured JSON with the requested URL, HTTP status, failed requests, HTTP failures, page errors, console errors/warnings, current `$world` state, and edit proof details. It fails fast on page errors, failed boot-critical JS/CSS/document requests, HTTP failures, console errors, or a captured world-transition error. It treats aborted XHR/fetch requests during world transitions as non-fatal because those can be normal when the loading screen hands off to the real world.

The edit proof loads `lively.morphic` and `lively.graphics`, adds a label morph to `$world`, updates its value, and verifies that the morph is owned by the world with the updated value. Load routes are not considered ready while the loading-screen morph is still present.

## What changed for debugging reliability

- `scripts/check_boot.js` now reports structured diagnostics and distinguishes dashboard visibility checks from world/project load readiness.
- `lively.freezer/src/util/bootstrap.js` records transition failures on `window.__loadError__` so the headless check can report the actual boot error instead of timing out on a blank/loading page.
- The JS linter install is now opt-in through `?installLinter=true` and loaded dynamically, so editor tooling dependencies do not block normal world/project boot.
- The unified freezer build uses `globalThis` as the Rollup context, keeps Babel as the final browser compatibility pass, and externalizes boot-irrelevant tooling/storage/parser packages that should not be part of the initial browser bundle.
- Browser bundling now lets Node built-ins fall through to browser polyfills, honors package browser remaps, provides a minimal `process.env` shim, and makes Rollup's missing-export shim callable for externalized boot modules.
- Fresh installs reset flatn package maps before importing newly installed dependencies such as `systemjs`, `@jspm/generator`, and `@babel/core`.
- Smaller compatibility fixes keep browser boot from being misclassified as Node, avoid a browser-breaking `semver` import in serializer checks, normalize Acorn's namespace/default wrapper shape, and tolerate an externalized `kld-intersections` module during startup.

## Verification

- `make hooks` attempted; Windows make resolved `pwd` incorrectly, so `core.hooksPath` was repaired to the repo `.githooks` path directly.
- `node --check` on changed JavaScript files.
- `bash -n install.sh`.
- `bash -n scripts/docker-entrypoint.sh`.
- `git diff --check upstream/main...HEAD`.
- `docker compose config`.
- Fresh disposable Docker Compose run from empty volumes.
- Dashboard boot proof through the container boot harness.
- World-load edit proof through the container boot harness.
- Named-volume state isolation check: generated directory contents were present inside Docker volumes while host mount-point directories stayed empty.
- GitHub `Check PR for merging` passed on the latest head.

## Notes

- Docker Compose is optional and separate from the native run-server path.
- The server is bound to `127.0.0.1:9011` for local development.
- Docker dependency/cache/database/build directory contents live in named volumes so they do not share platform-specific state with native installs.
- Project creation currently reaches the login/auth boundary; auth provisioning is intentionally left for a follow-up.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,7 @@
 .git
 
-# Host-visible generated dependency/cache directories are created through the
-# bind mount at runtime, not copied into the image build context.
+# Docker-owned generated dependency/cache directories are mounted or created at
+# runtime, not copied into the image build context.
 .puppeteer-browser-cache
 .livelydbs
 .module_cache
@@ -30,6 +30,8 @@ user-server.sh
 /lively.classes/build
 /lively.freezer/landing-page
 /lively.freezer/loading-screen
+/lively.freezer/swc-plugin/target
 /lively.installer.log
 /lively.server/usage_log.json
+/lively.server/.module_cache
 /mocha-es6/.cachedImportMap.json
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,35 @@
+.git
+
+# Host-visible generated dependency/cache directories are created through the
+# bind mount at runtime, not copied into the image build context.
+.puppeteer-browser-cache
+.livelydbs
+.module_cache
+components_cache
+custom-npm-modules
+dev-deps
+esm_cache
+lively.next-node_modules
+local_projects
+resources
+snapshots
+subservers
+tmp
+uploads
+users
+*.idb
+
+# Local runtime files are seeded by the entrypoint when absent.
+favicon.ico
+localconfig.js
+package-lock.json
+robots.txt
+user-server.sh
+/flatn/.cachedImportMap.json
+/lively.*/.cachedImportMap.json
+/lively.classes/build
+/lively.freezer/landing-page
+/lively.freezer/loading-screen
+/lively.installer.log
+/lively.server/usage_log.json
+/mocha-es6/.cachedImportMap.json
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,21 @@
-* text=auto
+# This repo is built and validated primarily on Linux runners. Keep text files
+# checked out with LF so shell scripts, generated browser bundles, and source
+# patches behave the same on Windows, macOS, and Linux.
+* text=auto eol=lf
+
+# Keep common binary assets out of line-ending normalization.
+*.br binary
+*.gif binary
+*.gz binary
+*.ico binary
+*.jpg binary
+*.jpeg binary
+*.otf binary
+*.pdf binary
+*.png binary
+*.ttf binary
+*.wasm binary
+*.webp binary
+*.woff binary
+*.woff2 binary
+*.zip binary
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,106 @@
+FROM ubuntu:24.04
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+# Ubuntu 24.04 is pinned so the development runtime stays reproducible while
+# still matching a current LTS release supported by NodeSource's Node 24 setup.
+ENV DEBIAN_FRONTEND=noninteractive \
+    LANG=C.UTF-8 \
+    LC_ALL=C.UTF-8 \
+    BUN_INSTALL=/root/.bun \
+    RUSTUP_HOME=/opt/rustup \
+    CARGO_HOME=/opt/cargo
+
+ENV PATH="${CARGO_HOME}/bin:${BUN_INSTALL}/bin:${PATH}"
+
+# Install the base OS tools, build chain, browser shared libraries, and fonts
+# needed for a host with only Docker to run install.sh, start the server, and
+# execute Puppeteer-based boot checks inside the container.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+      apt-transport-https \
+      aspell \
+      bash \
+      brotli \
+      build-essential \
+      ca-certificates \
+      curl \
+      dumb-init \
+      entr \
+      fonts-freefont-ttf \
+      fonts-ipafont-gothic \
+      fonts-kacst \
+      fonts-liberation \
+      fonts-thai-tlwg \
+      fonts-wqy-zenhei \
+      git \
+      gnupg \
+      iproute2 \
+      libasound2t64 \
+      libatk-bridge2.0-0 \
+      libatk1.0-0 \
+      libatspi2.0-0 \
+      libcairo2 \
+      libcups2 \
+      libdbus-1-3 \
+      libdrm2 \
+      libgbm1 \
+      libglib2.0-0 \
+      libgtk-3-0 \
+      libnspr4 \
+      libnss3 \
+      libpango-1.0-0 \
+      libx11-6 \
+      libx11-xcb1 \
+      libxcb1 \
+      libxcomposite1 \
+      libxdamage1 \
+      libxext6 \
+      libxfixes3 \
+      libxkbcommon0 \
+      libxrandr2 \
+      libxss1 \
+      lsb-release \
+      net-tools \
+      perl \
+      pkg-config \
+      procps \
+      python3 \
+      python3-pip \
+      python3-venv \
+      sudo \
+      tar \
+      tree \
+      unzip \
+      wget \
+      xdg-utils \
+      xz-utils \
+    && rm -rf /var/lib/apt/lists/*
+
+# NodeSource provides Node 24 packages for Noble; using the distro package keeps
+# node and npm available without requiring a host-level version manager.
+RUN curl -fsSL https://deb.nodesource.com/setup_24.x -o /tmp/nodesource_setup.sh \
+    && bash /tmp/nodesource_setup.sh \
+    && apt-get install -y --no-install-recommends nodejs \
+    && rm -f /tmp/nodesource_setup.sh \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN curl -fsSL https://bun.sh/install | bash
+
+# Rust is installed in a shared location because the project build needs the
+# wasm32-wasip1 target and the container may run as different users later.
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
+      | sh -s -- -y --profile minimal --default-toolchain stable \
+    && rustup target add wasm32-wasip1 \
+    && chmod -R a+rwx "${RUSTUP_HOME}" "${CARGO_HOME}"
+
+RUN pip3 install --break-system-packages --no-cache-dir sultan
+
+WORKDIR /workspace
+
+COPY scripts/docker-entrypoint.sh /usr/local/bin/lively-docker-entrypoint
+RUN chmod +x /usr/local/bin/lively-docker-entrypoint
+
+EXPOSE 9011
+
+ENTRYPOINT ["dumb-init", "--", "lively-docker-entrypoint"]
--- a/README.md
+++ b/README.md
@@ -55,7 +55,21 @@ For some more advanced development operations (such as bulk testing from the com
 >
 > If you want to change files outside of `lively` (i.e., in a normal editor), while still having the changes be available in lively when opening the file, you'll need to install `entr` from its [repository](https://github.com/eradman/entr). Usually, when working inside of `lively.next`, this will not be an issue, but it can be handy when working heavily on the core of `lively`. *This feature works semi-reliable at the moment. If you are interested in this and would like to help debug this, please reach out!*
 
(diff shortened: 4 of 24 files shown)


tashabits:
## Contribution guide cleanup note

I did a draft cleanup pass to better align this PR with `CONTRIBUTING.md` before review:

- rewrote the branch into three maintainer-facing commits grouped by runtime setup, browser boot hardening, and documentation
- updated the PR title/body to match the larger-change pattern used in recent upstream PRs
- confirmed the local hook configuration for this checkout; `make hooks` hit a Windows `pwd` issue, so `core.hooksPath` was set directly to the repo `.githooks` path
- reran syntax/static checks and a warm dashboard boot smoke after the rewrite

The PR is intentionally still draft while final maintainer-facing review polish happens.

merryman:
Does this address the issues that made us deprecate our previous docker setup? https://github.com/LivelyKernel/lively.next/pull/922

One of these were for instance that instantiating the database in docker made it impossible to later consume that database from MacOS and vice versa. A similar issue was with the installed node modules in `lively.next-node_modules` which would work on either the host system or docker but not both. So has this solution figured out a way around this issue?

tashabits:
> Does this address the issues that made us deprecate our previous docker setup? #922
> 
> One of these were for instance that instantiating the database in docker made it impossible to later consume that database from MacOS and vice versa. A similar issue was with the installed node modules in `lively.next-node_modules` which would work on either the host system or docker but not both. So has this solution figured out a way around this issue?

Both yes and no.

#1799 makes Docker a real, self-contained dev launcher. The host no longer needs Node/Bun/Rust/Python/browser deps installed, and Compose gives it a stable entrypoint instead of the old Makefile-run-container workflow from #922.

It does, however, still bind-mount the checkout and writes generated state into host-visible dirs like lively.next-node_modules, .livelydbs, snapshots, and esm_cache. That means Docker/Linux and native macOS can still share platform-sensitive artifacts.

If you want those fixed as an isolation fix where docker uses named volumes,  the entrypoint/install adjusted to check those docker-owned volumes and the docker state is distinct from the native state, that's easy enough to do.  It's a much bigger job to make the *same* database genuinely visible and usable from both the host and container, it's way harder.

merryman:
> > Does this address the issues that made us deprecate our previous docker setup? #922
> > One of these were for instance that instantiating the database in docker made it impossible to later consume that database from MacOS and vice versa. A similar issue was with the installed node modules in `lively.next-node_modules` which would work on either the host system or docker but not both. So has this solution figured out a way around this issue?
> 
> Both yes and no.
> 
> #1799 makes Docker a real, self-contained dev launcher. The host no longer needs Node/Bun/Rust/Python/browser deps installed, and Compose gives it a stable entrypoint instead of the old Makefile-run-container workflow from #922.
> 
> It does, however, still bind-mount the checkout and writes generated state into host-visible dirs like lively.next-node_modules, .livelydbs, snapshots, and esm_cache. That means Docker/Linux and native macOS can still share platform-sensitive artifacts.
> 
> If you want those fixed as an isolation fix where docker uses named volumes, the entrypoint/install adjusted to check those docker-owned volumes and the docker state is distinct from the native state, that's easy enough to do. It's a much bigger job to make the _same_ database genuinely visible and usable from both the host and container, it's way harder.

If you manage to cleanly separate these by named volumes I think we are fine. Unless @linusha has some other concerns I have not thought about yet.

tashabits:
> > > Does this address the issues that made us deprecate our previous docker setup? #922
> > > One of these were for instance that instantiating the database in docker made it impossible to later consume that database from MacOS and vice versa. A similar issue was with the installed node modules in `lively.next-node_modules` which would work on either the host system or docker but not both. So has this solution figured out a way around this issue?
> > 
> > 
> > Both yes and no.
> > #1799 makes Docker a real, self-contained dev launcher. The host no longer needs Node/Bun/Rust/Python/browser deps installed, and Compose gives it a stable entrypoint instead of the old Makefile-run-container workflow from #922.
> > It does, however, still bind-mount the checkout and writes generated state into host-visible dirs like lively.next-node_modules, .livelydbs, snapshots, and esm_cache. That means Docker/Linux and native macOS can still share platform-sensitive artifacts.
> > If you want those fixed as an isolation fix where docker uses named volumes, the entrypoint/install adjusted to check those docker-owned volumes and the docker state is distinct from the native state, that's easy enough to do. It's a much bigger job to make the _same_ database genuinely visible and usable from both the host and container, it's way harder.
> 
> If you manage to cleanly separate these by named volumes I think we are fine. Unless @linusha has some other concerns I have not thought about yet.

This little patch should address the main concern from #922 by keeping Docker’s generated state separate from native runs. Source files are still bind-mounted for editing, but dependency/db/cache/build directories now live in Docker named volumes, so Linux container artifacts don’t get mixed into a macOS/native checkout.

It also uses `nocopy: true` on those volumes, so a newly created Docker volume is not seeded from stale host artifacts that might already exist in the checkout.

linusha:
Just a quick chime-in from my side: I haven't looked at everything in real detail, but I think the important part is solely acknowledging the differences/(non)-compatibilities in the README, which this PR does an excellent job at!

@merryman I am not 100% sure anymore, but I think the original motivation for docker support was that we thought that might be a viable solution to get stuff to build on windows at the time. Later we just suggested WSL. Could that be? One of the reasons being that some deps that were compiled were always either compatible with the native host or the docker one, which would also be the case here, but they wouldn't interfere with each other, so it does not matter that.

I don't have the resources to stay anything substantial code-wise, but the thoughts that I would have on principal are nicely addressed here! Thanks for your work @tashabits!

tashabits:
@linusha @merryman I think the next natural step on this branch's path would be to build out an option where you could create a pinned runnable container that users could download and play with.

The current Compose setup is shaped well for development: it bind-mounts the checkout as /workspace and keeps generated Docker artifacts in named volumes so they don’t collide with host/native installs. But that also means it is not really a distributable “working object.”

A sibling runtime image could be different: copy the repo into the image, run the long install/build path at image build time, and start directly into `./start-server.sh 9011`

You could even run it on a schedule so it keeps publishing runnable versions to a public container registry if there have been changes, if you wanted.

Happy to add that as part of this branch too if it feels like the right scope: a separate runtime Dockerfile/Compose path for the pinned image, while keeping the current Compose setup focused on local development.

tashabits:
Added a fix in `c831fc8a4` for the MorphicDB boot path exposed by the compose volume/freezer setup.

What changed:
- `lively.morphic/morphicdb/db.js` now uses a small MorphicDB-local HTTP interface for ObjectDB endpoints.
- This avoids importing `lively.storage`/`ObjectDBHTTPInterface` in the browser boot path, which was pulling the storage/PouchDB dependency graph into the freezer world path after the Docker volume changes.

Validation run:
- `docker compose down -v --remove-orphans`
- `docker compose up --build -d`
- waited for `server sucessfully started` and `finished library snapshot`
- loaded `http://localhost:9011/worlds/load?name=__newWorld__&askForWorldName=false&fastLoad=false`
- verified `$world.isNotUnique('CodexCleanStackWorldName') === false`
- dragged two rectangles into the world
- opened a halo/Object Editor path
- saved `onMouseDown(evt)` handlers through Object Editor with `super.onMouseDown(evt)`
- clicked both morphs in Hand mode; the second handler looked up the first morph by name and logged its position, extent, fill, and property keys

The clean-volume compose stack is currently up on `127.0.0.1:9011`.

merryman (commented):
There are some major issues with this branch. Commits 70300d2fd9ded038466721cce236442a1cc10235, e0bf1e6fbe8425d513ee47175fa51dedf41c9ea6, 2a8aad400f1fa690db0cc76407c9ca0d46d8444, 22970836324e177211cc2d5927c7aa588330bda4, 754101639cd2c173b3c333de3c46177fa1e29d53 and c831fc8a41da24b1206e785b759d2c88c99040b5. All are hacky workarounds for dealing with broken import maps. The only commit that actually tries to adress problems with import maps is baa9967da36fe6be208028b7c89d9de13ce8a47d, but its not entirely clear what exactly it adresses. Can you clarify the issue you ran into with import maps? On main the import map appear to work fine (the daily check as of now passes meaning all the tests run fine).

merryman (changes requested):

tashabits:
> There are some major issues with this branch. Commits [70300d2](https://github.com/LivelyKernel/lively.next/commit/70300d2fd9ded038466721cce236442a1cc10235), [e0bf1e6](https://github.com/LivelyKernel/lively.next/commit/e0bf1e6fbe8425d513ee47175fa51dedf41c9ea6), [2a8aad4](https://github.com/LivelyKernel/lively.next/commit/2a8aad400f1fa690db0cc76407c9ca0d46d84441), [2297083](https://github.com/LivelyKernel/lively.next/commit/22970836324e177211cc2d5927c7aa588330bda4), [7541016](https://github.com/LivelyKernel/lively.next/commit/754101639cd2c173b3c333de3c46177fa1e29d53) and [c831fc8](https://github.com/LivelyKernel/lively.next/commit/c831fc8a41da24b1206e785b759d2c88c99040b5). All are hacky workarounds for dealing with broken import maps. The only commit that actually tries to adress problems with import maps is [baa9967](https://github.com/LivelyKernel/lively.next/commit/baa9967da36fe6be208028b7c89d9de13ce8a47d), but its not entirely clear what exactly it adresses. Can you clarify the issue you ran into with import maps? On main the import map appear to work fine (the daily check as of now passes meaning all the tests run fine).

@merryman I messaged you on slack and I'll give you a thorough write-up of what I was encountering and what all those commits were trying to do

tashabits:
I will open another PR when I address those, but committing now will just be noisy and waste CI minutes...I'll get another to you when I have some time to put to it

